PR #7, opened 2018-07-30
Finish title conversion with specs

mejackreed:
~~I've left some follow on todo's here and ported specs and marked them as pending.~~

The todos should all be done. One remaining pending is due to a difference in punctuation trimming logic.


lib/traject/config/sirsi_config.rb
```diff
@@ -33,22 +33,51 @@
...
+      str = field['k']
+    end
+    if str.nil?
+      # still? All we can do is bail, I guess
+      return nil
+    end
+
+    non_filing = field.indicator2.to_i
+    str = str.slice(non_filing, str.length)
+    str = str.delete(java7_regex_punct).delete('$').strip
```
cbeer: I think the `.delete('$')` is now covered by the special java punctuation list (which isn't a regex, so maybe the variable should be renamed?)

mejackreed: Good point, I've fixed it up now.

PR #24, opened 2018-07-31
select first record only for these non multivalued fields

### LINKED ISSUE #23: Address multivalued / singlevalued errors
When indexing our real data using sw solr confs, getting errors like:

```
2018-07-31T07:27:25-06:00  INFO    Indexer with 1 processing threads, reader: Traject::MarcReader and writer: Traject::SolrJsonWriter
2018-07-31T07:27:46-06:00 ERROR Error in Solr batch add. Will retry documents individually at performance penalty: Solr response: 400: {"responseHeader":{"status":400,"QTime":3},"error":{"metadata":["error-class","org.apache.solr.common.SolrException","root-error-class","org.apache.solr.common.SolrException"],"msg":"ERROR: [doc=21764] multiple values encountered for non multiValued field title_245_search: [Hyōjun Nihongo tokuhon : [The standard Japanese readers] /, 標準日本語讀本 [The standard Japanese readers]]","code":400}}
```

mejackreed:
Fixes #23 

~~I'm not sure if we should be joining these fields instead of selecting the first one?~~ Now testing against current sw behavior.


cbeer:
Should we see what the current behavior is? Last time I did it, I just spun up a searchworks solr collection locally with all the fields as `stored=true`.

mejackreed:
I went back here and indexed the records using solrmarc-sw (all stored fields) and then created feature specs based on errors encountered and previous behavior.

PR #33, opened 2018-08-01
add initial benchmark, hopefully to let us know if we do something bad

mejackreed:


mejackreed:
Should we allow failures on travis? or do something else?

PR #52, opened 2018-08-06
allow travis to fail on jruby (bug in jruby that implements slice_whe…

mejackreed:
…n incorrectly)


cbeer:
or #53; pick the lesser of two evils.

PR #53, opened 2018-08-07
Add a jruby-compatible way to combine records

cbeer:
Workaround for https://github.com/jruby/jruby/issues/5275.


mejackreed:
Or the third option to monkey patch `Enumerable` (for jruby only)?

lib/traject/readers/marc_combining_reader.rb
```diff
@@ -1,8 +1,41 @@
...
+  end
+end
+
+class Traject::MarcCombiningReader
+  include Enumerable
+
+  attr_reader :settings, :input_stream, :marc_reader
+
+  def initialize(input_stream, settings)
+    @marc_reader = Traject::MarcReader.new(input_stream, settings)
```
mejackreed: Does this then not use the `traject-marc4j_reader`? Are there side effects of this wrt performance?

cbeer: Perhaps, but adding support for marc4j reader is blocked by https://github.com/traject/traject-marc4j_reader/pull/2.

PR #54, opened 2018-08-07
Excludes 1xx relator term subfields from author_sort.

shelleydoljack:


cbeer:
Potentially blocked, waiting for #47.

PR #58, opened 2018-08-07
add course reserves fields

### LINKED ISSUE #27: Map CREZ-indexed fields
- [x] crez_instructor_search
- [x] crez_course_name_search
- [x] crez_course_id_search
- [x] crez_desk_facet
- [x] crez_dept_facet
- [x] crez_course_info

And it modifies `item_display` and `building_facet`

See https://github.com/sul-dlss/crez-sw-ingest/blob/master/lib/add_crez_to_solr_doc.rb

mejackreed:
Fixes #27


lib/traject/config/sirsi_config.rb
```diff
@@ -13,6 +14,24 @@
   provide "reader_class_name", "Traject::MarcCombiningReader"
 end
 
+RESERVES_FILE = ENV['RESERVES_FILE'] || './spec/fixtures/files/multmult.csv'
```
cbeer: I wonder if this can be a traject setting instead of an env variable directly? That might make it easier to test (and, I assume, mean we won't need to hard-code a spec-path location here?)

mejackreed: k. what do you think about this approach?

PR #107, opened 2018-08-10
Fixes to the author_sort field

cbeer:


lib/traject/config/sirsi_config.rb
```diff
@@ -246,7 +246,7 @@ def extract_sortable_title(fields, record)
 # # Author Sort Field
 to_field 'author_sort' do |record, accumulator|
   accumulator << extract_sortable_author('100abcdefgjklnpqtu:110abcdefgklnptu:111acdefgjklnpqtu',
-                                         '240adfghklmnoprs:245abcfghknps',
+                                         '240acdfghklmnoprs:245abfghknps',
```
shelleydoljack: 240 $c is not a valid subfield. I found we have 6 records with 240 $c, but I think they are errors. I will ask Data Control regarding those records.

lib/traject/config/sirsi_config.rb
```diff
@@ -260,7 +260,7 @@ def extract_sortable_title(fields, record)
 #  ensures record with no 1xx sorts after records with a 1xx by prepending UTF-8 max code point to title string
 def extract_sortable_author(author_fields, title_fields, record)
   punct = '!"#$%&\'()*+,-./:;<=>?@[\]^_`{|}~\\'
-  onexx = Traject::MarcExtractor.cached(author_fields).collect_matching_lines(record) do |field, spec, extractor|
+  onexx = Traject::MarcExtractor.cached(author_fields, alternate_script: false).collect_matching_lines(record) do |field, spec, extractor|
```
shelleydoljack: Good catch.

lib/traject/config/sirsi_config.rb
shelleydoljack: Data Control confirmed that the records with 240 $c were mistakes. But as you can see, it's possible for records to have a 240 $c. I think we should exclude it from the author_sort.

cbeer: The solrmarc code currently includes them; what do you think about making a ticket to remove this behavior after we can verify we produce identical data?

shelleydoljack: Ok. #110

PR #114, opened 2018-08-13
Remove extra code from building_facet

camillevilla:
😬


mejackreed:
Is there a quick test we can add for this?

cbeer:
I think this is just a pure refactor, with no visible changes, no?

PR #122, opened 2018-08-15
Fixes

cbeer:


lib/traject/config/sirsi_config.rb
```diff
@@ -104,6 +104,22 @@ def open_collection(use_ns)
   accumulator << result.join(' ') if result.any?
 end
 
+class Traject::MarcExtractor
+  def collect_subfields(field, spec)
```
mejackreed: Could you add a quick comment about this monkey patch?

cbeer: 👍

PR #148, opened 2018-08-17
Takes only first title field solr schema expects to be not-repeatable.

### LINKED ISSUE #108: remove linked fields from title search
I think we need to do `alternate_script: false` for these title search fields since there are already vern_title*_search fields.
```
to_field 'title_245a_search', extract_marc('245a', first: true)
to_field 'vern_title_245a_search', extract_marc('245a', alternate_script: :only)
to_field 'title_245_search', extract_marc('245abfgknps', first: true)
to_field 'vern_title_245_search', extract_marc('245abfgknps', alternate_script: :only)
```

shelleydoljack:
Closes #108


spec/lib/traject/config/title_spec.rb
```diff
@@ -54,16 +64,35 @@
       result = select_by_id('245nAndp')[field]
       expect(result.first).to include 'humanities'
     end
+    context 'linked fields' do
+      let(:fixture_name) { 'vernacularSearchTests.mrc' }
+      it 'are not included' do
+        expect(select_by_id('2xxVernSearch')[field]).not_to eq ['vern245a',
```
mejackreed: With the below expectation `[field]).to eq ` can we get away with out adding this test?

PR #149, opened 2018-08-17
Crez updates to item_display and building_facet

### LINKED ISSUE #59: Modify crez fields
Crez indexing modifies `item_display` and `building_facet`

See https://github.com/sul-dlss/crez-sw-ingest/blob/master/lib/add_crez_to_solr_doc.rb

Prerequisites: #22

A spin off ticket from #27

camillevilla:
Fixes #59


lib/translation_maps/library_code_translations.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+{
```
cbeer: This seems suspiciously like our `library_map.properties` file?

mejackreed: Yep. Looks like there are two minor differences. Should we consolidate now?

mejackreed: Though I wonder if the "lack of" the 13 additional keys at the end could cause problems.. I think that is why we just ported over what we had for now.

cbeer: Ah, I didn't scroll down to the bottom. I'll push the green button and let @jkeck chime in next week.

PR #152, opened 2018-08-21
Update LC regex to allow some weird lane callnumbers to go through

cbeer:


lib/sirsi_holding.rb
```diff
@@ -99,7 +99,7 @@ def physics_not_temp?
   class CallNumber
     BEGIN_CUTTER_REGEX = /( +|(\.[A-Z])| *\/)/
     VALID_DEWEY_REGEX = /^\d{1,3}(\.\d+)? *\.? *[A-Z]\d{1,3} *[A-Z]*+.*/
-    VALID_LC_REGEX = /(^[A-Z&&[^IOWXY]]{1}[A-Z]{0,2} *\d+(\.\d+)?( +([\da-z]\w*)|([A-Z]\D+[\w]*))?) *\.?[A-Z]\d+.*/
+    VALID_LC_REGEX = /(^[A-Z&&[^IOWXY]]{1}[A-Z]{0,2} *\d+\.?(\d+)?( +([\da-z]\w*)|([A-Z]\D+[\w]*))?) *\.?[A-Z]\d+.*/
```
jkeck: Should this be `*\d+(\.?\d+)?` instead of `*\d+\.?(\d+)?` (keeping the `period` in the capture group), or maybe it doesn't matter?

cbeer: I was trying to match `R711. P7`. I guess we need it as part of the capture group for later, though, huh?

Issue #143, opened 2018-08-16
Resolve jruby in prod issues


mejackreed:

mejackreed:
Also seeing this from the traject process:

```
[harvestdor@sw-indexing-dev current]$ SOLR_URL=https://sul-solr.stanford.edu/solr/sw-traject-1 NUM_THREADS=0 bundle exec traject -c lib/traject/config/sirsi_config.rb -s solr_writer.max_skipped=-1 -s log.file=/data/sirsi/latest/logs/mark-x.log /data/sirsi/latest/uni_00000000_00499999.marc
bundler: failed to load command: traject (/opt/app/harvestdor/searchworks_traject_indexer/shared/bundle/jruby/2.5.0/bin/traject)
LoadError: no such file to load -- /opt/app/harvestdor/searchworks_traject_indexer/shared/bundle/jruby/2.5.0/gems/traject-2.3.4/bin/traject
  org/jruby/RubyKernel.java:994:in `load'
  /opt/app/harvestdor/searchworks_traject_indexer/shared/bundle/jruby/2.5.0/bin/traject:23:in `<main>'
  org/jruby/RubyKernel.java:994:in `load'
```
I wonder if its because I'm setting rvm just for the shell?

mejackreed:
Re running with `.ruby-version` file manually setting jruby on the server.

mejackreed:
Things to try:

 - [ ] set jruby as default ruby and run
 - [ ] fork traject and add in additional catch

mejackreed:
Full stacktrace:

```
2018-08-20T08:48:01-07:00 ERROR Error in Solr batch add. Will retry documents individually at performance penalty: Solr response: :
2018-08-20T08:48:01-07:00 FATAL Traject::CommandLine: Unexpected exception, terminating execution: java.net.SocketException: Connection closed by remote host
2018-08-20T08:48:01-07:00 FATAL sun.security.ssl.SSLSocketImpl.checkWrite(sun/security/ssl/SSLSocketImpl.java:1565)
	sun.security.ssl.AppOutputStream.write(sun/security/ssl/AppOutputStream.java:71)
	java.io.OutputStream.write(java/io/OutputStream.java:75)
	java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:498)
	org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(org/jruby/javasupport/JavaMethod.java:455)
[...]

PR #154, opened 2018-08-21
temporarily fixes traject in jruby prod issues until https://github.c…

mejackreed:
…om/traject/traject/pull/194 gets merged and released

Part of #143


mejackreed:
Superceded by #157 and no longer needed as long as we don't start throwing crazy errors.

PR #157, opened 2018-08-23
 Add a Manticore HTTP client for jruby

cbeer:


lib/traject/config/sirsi_config.rb
```diff
@@ -26,6 +26,12 @@
   provide 'reserves_file', ENV['RESERVES_FILE']
   provide 'allow_duplicate_values',  false
   provide 'skip_empty_item_display', ENV['SKIP_EMPTY_ITEM_DISPLAY'].to_i
+  provide 'solr_json_writer.http_client', Traject::HttpRbHttpClient.new
```
mejackreed: Cruft?

cbeer: Yep; rebasing is hard.

mejackreed: Yeah.. I did the _exact_ same thing yesterday

PR #160, opened 2018-08-24
 Add a PurlFetcherReader to stream druids to index directly from purl-fetcher

cbeer:


lib/traject/readers/purl_fetcher_reader.rb
```diff
@@ -0,0 +1,74 @@
...
+  ##
+  # @return [Enumerator]
+  def deletes(params = {})
+    paginated_get('/docs/deletes', 'deletes', params)
+  end
+
+  ##
+  # @return [Hash] a parsed JSON hash
+  def get(path, params = {})
+    JSON.parse(HTTP.get("https://purl-fetcher-prod.stanford.edu" + path, params: params).body)
```
cbeer: If we go this way, this should probably be configurable in a way we don't yet have available to us.

lib/traject/readers/purl_fetcher_reader.rb
```diff
@@ -0,0 +1,74 @@
...
+  ##
+  # @return [Enumerator]
+  def deletes(params = {})
+    paginated_get('/docs/deletes', 'deletes', params)
+  end
+
+  ##
+  # @return [Hash] a parsed JSON hash
+  def get(path, params = {})
+    JSON.parse(HTTP.get(settings.fetch('purl_fetcher.api_endpoint', 'https://purl-fetcher-prod.stanford.edu') + path, params: params).body)
```
mejackreed: Do we need `-prod` here?

cbeer:
Yeah, I'll add a test if this is the approach we want to take.

PR #188, opened 2018-09-13
Add whenever

cbeer:


script/index_sirsi_full.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+set -e
+
+if [[ $(date "+%u") -eq 0 ]];
```
cbeer: I hope we can move this check either to a wrapper script or put it behind a flag? It'd be really annoying if we can only ever build a new index on a Sunday.

PR #224, opened 2018-10-05
Add Mods display fields

mejackreed:
This PR adds new fields needed to support no mods parsing on search results display in https://github.com/sul-dlss/SearchWorks/issues/1987

### DIFF
--- a/lib/traject/config/sdr_config.rb
+++ b/lib/traject/config/sdr_config.rb
@@ -40,6 +40,18 @@ def mods_xpath(xpath)
   end
 end
 
+def mods_display(method, *args, default: nil)
+  lambda do |resource, accumulator, _context|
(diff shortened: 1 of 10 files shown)


lib/traject/config/sdr_config.rb
```diff
@@ -34,6 +34,18 @@ def stanford_mods(method, *args, default: nil)
   end
 end
 
+def mods_display(method, *args, default: nil)
+  lambda do |resource, accumulator, _context|
+    data = Array(resource.mods_display.public_send(method, *args))
+
+    data.each do |v|
+      accumulator << v.values.first.to_s
```
shelleydoljack: From Arcadia’s document for mods indexing parity, there is a desire to use the attribute usage=primary before taking the first name element.

PR #225, opened 2018-10-05
Index marc data structured used for record display on search results

cbeer:
This is the indexing part of https://github.com/sul-dlss/SearchWorks/issues/1988


lib/traject/config/sirsi_config.rb
```diff
@@ -312,6 +395,35 @@ def extract_sortable_title(fields, record)
...
+  struct[:creator] = linked_author_struct(record, '100')
+  struct[:corporate_author] = linked_author_struct(record, '110')
+  struct[:meeting] = linked_author_struct(record, '111')
+  struct.reject! { |_k, v| v.empty? }
+
+  accumulator << struct unless struct.empty?
+end
+
+def linked_author_struct(record, tag)
+  record.fields(tag).map do |field|
```
cbeer: Ugh; it looks like we'll have to implement the Searchworks MarcField `run_preprocessors` steps too..

PR #230, opened 2018-10-15
[WIP] Increases thread pools for writing docs to solr.

shelleydoljack:


shelleydoljack:
https://github.com/sul-dlss/sul-solr-configs/pull/174 seems to have fixed the timeout issues.

PR #245, opened 2018-10-24
Adds switch for indexing all MODS.

shelleydoljack:


script/index_sdr.sh
```diff
@@ -5,10 +5,16 @@ STATE_FILE=tmp/searchworks_traject_indexer_last_run
 CURRENT_DATE=`date --rfc-3339=seconds`
 export NUM_THREADS=8
 LOG_FILE=tmp/index_sdr_log
+ALL_MODS=$1
+
+if [ $ALL_MODS ]; then #checks if ALL_MODS is defined, else index only latest
+  LAST_DATE="2000-01-01 00:00:05-07:00"
```
cbeer: Does `$LAST_DATE` need a value, or could we just initialize it to an empty string?

cbeer:
Closing; this script is gone. Now we'd do something like:

```
/opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092  --group $KAFKA_TOPIC  ----reset-offsets --to-earliest --execute
```

PR #252, opened 2018-11-01
add iiif manifest url for appropriate content types

### LINKED ISSUE #243: IIIF badge - Indexing tasks
We need to identify which records should get an IIIF badge (see comment below) and add the manifest URL to a new field in the document.

We'll ultimately want to be able to query this in the app to create the resource access point. So either we need to be able to do some sort of query against the field to return documents w/ any value populated, or have some sort of facetable value (whichever makes sense)

mejackreed:
Fixes #243 

Adds both SDR and SIRSI iiif manifests


lib/traject/config/sdr_config.rb
```diff
@@ -331,6 +331,12 @@ def mods_display(method, *args, default: nil)
 
 to_field 'summary_display', mods_display(:abstract)
 
+to_field 'iiif_manifest_url_ssim' do |record, accumulator|
+  if %w[image manuscript map book].include?(record.dor_content_type)
```
cbeer: The logic in purl for whether an object has IIIF material is a little different:

https://github.com/sul-dlss/purl/blob/master/app/models/iiif_presentation_manifest.rb#L21-L26

Do these need to be reconciled at all?

mejackreed: I guess is the outstanding case here where content is a book, but doesn't have pages/images?

mejackreed:
@jkeck added some logic here: 0398f0dcd705bafa2be35bb36d8ec40f57c03c3d

Not sure if that is comprehensive enough though.. any thoughts on a better way?

lib/traject/config/sirsi_config.rb
```diff
@@ -3023,6 +3023,24 @@ def library_has(field)
...
+to_field 'iiif_manifest_url_ssim' do |record, accumulator|
+  Traject::MarcExtractor.new('856x').collect_matching_lines(record) do |field, spec, extractor|
+    subfields = extractor.collect_subfields(field, spec)
+    next unless subfields[0] == 'SDR-PURL' && subfields[1] == 'item'
+
+    accumulator.concat(subfields.slice(2..-1).map do |v|
+      v.split(':')
+    end.select do |(type, file_id)|
+      type == 'file' && /.*\.jp2/.match(file_id)
+    end.map do |(_type, file_id)|
```
jkeck: I think this is fine.  I'm not exactly sure what we would do w/ multiple IIIF manifests in the same doc, but not really a concern of this code.

lib/traject/config/sdr_config.rb
mejackreed: I think we should update this logic in PURL to be consistent with this logic.

PR #258, opened 2018-11-08
Fix toc_struct; the fields information is arrays-of-arrays

cbeer:


lib/traject/config/sirsi_config.rb
```diff
@@ -1848,7 +1848,11 @@ def marc_008_date(byte6values, byte_range, u_replacement)
 to_field "toc_search", extract_marc("905art:505art", alternate_script: false)
 to_field "vern_toc_search", extract_marc("505art", alternate_script: :only)
 
-# Generate dt/dd pair with an unordered list from the table of contents (IE marc 505s)
+# Generate structured data from the table of contents (IE marc 505 + 905s).
+# There are arrays of values for each TOC entry, and each TOC contains e.g. a chapter title; e.g.:
+#  - fields: [['Vol 1 Chapter 1', 'Vol 1 Chapter 2'], ['Vol 2 Chapter 1']]
+#    vernacular: [['The same, but pulled from the matched vernacular fields']]
+#    unmatched_vernacular: [['The same, but pulled from any unmatched vernacular fields']]
```
cbeer: I've also noticed the SearchWorks UI doesn't care about the distinction we're making between these, so it's possible to refactor it to just be a simple multivalued field, with each value containing an array of the chapters. I'm not sure I'm eager to take that on just now, though..

jkeck: I'm guessing this is a hold-over from the way SW used to handle TOCs.  We can prob write a tech-debt ticket to clean this up, but I share the ambivalence about dealing w/ it now.

PR #272, opened 2018-11-21
Convert indexers into daemons

cbeer:


cbeer:
Exactly; cron is only used for queueing (and hopefully that'll change when we can stream changes from symphony or do something better than purl_fetcher). There's no change to parallel processing for marc, and this provides the structure for parallel processing MODS.

PR #289, opened 2018-11-30
Store and deduplicate messages to guard against putting the same reco…

cbeer:
…rd on the pile multiple times

### DIFF
--- a/lib/traject/extractors/crez_marc_kafka_extractor.rb
+++ b/lib/traject/extractors/crez_marc_kafka_extractor.rb
@@ -10,12 +10,20 @@ class Traject::CrezMarcKafkaExtractor
     @topic = topic
   end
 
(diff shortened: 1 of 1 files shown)


lib/traject/extractors/crez_marc_kafka_extractor.rb
```diff
@@ -10,12 +10,20 @@ def initialize(reserves_file:, kafka:, topic:)
     @topic = topic
   end
 
+  # Scan through the Kafka topic and re-send any messages for records with CREZ data
   def process!
     t0 = Time.now
+    h = {}
```
cbeer: There may be some very clever way to read the topic backwards so we don't need to store a bunch of records all the time, but hopefully this whole file is very temporary and will go away within the month..

camillevilla: 🤞

PR #368, opened 2019-08-07
Proof of concept indexing non-georeferenced images/maps into EarthWorks

mejackreed:


cbeer:
This seems like it might be ready to ship, but we only wire it up on -stage?

mejackreed:
I have a few more commits to push after pairing with @kimdurante on it. I'll ping this thread back when ready.

mejackreed:
@cbeer I've marked this as ready

PR #382, opened 2019-10-01
Add catch-some fields for key bib data

cbeer:


jcoyne:
@cbeer is the feature flag comment still relevant?  I didn't spot that.

cbeer:
Nope, that was only in the original PR from 2019 😬

jcoyne (approved):

PR #388, opened 2019-10-17
Update regex for TOC to require either a colon, period, or hyphen...

### LINKED ISSUE #387: Content notes (505) should be line breaking at --
examples 
https://searchworks.stanford.edu/view/12845889
https://searchworks.stanford.edu/view/13312935

SW-3854
SW-3848

jkeck:
…er a number before using it to split on.

Fixes #387

See https://github.com/sul-dlss/searchworks_traject_indexer/issues/387#issuecomment-542946027 for some more context/details.

### DIFF
--- a/lib/traject/config/sirsi_config.rb
+++ b/lib/traject/config/sirsi_config.rb
@@ -1918,7 +1918,13 @@ to_field "vern_toc_search", extract_marc("505art", alternate_script: :only)
 #    vernacular: [['The same, but pulled from the matched vernacular fields']]
 #    unmatched_vernacular: [['The same, but pulled from any unmatched vernacular fields']]
 to_field 'toc_struct' do |marc, accumulator|
-  formatted_chapter_regex = Regexp.union(/[^\S]--[^\S]/, /      /, /(?=(?:Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i,  /(?=(?<!Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i, /(?=(?:Appendix|Section|Chapter) [XVI]+[\.-]?)/i)
+  formatted_chapter_regex = Regexp.union(
+    /[^\S]--[^\S]/,
+    /      /,
+    /(?=(?:Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i,
+    /(?=(?<!Chapter|Section|Appendix|Part|v\.) \d+[:\.-]\s+)/i,
+    /(?=(?:Appendix|Section|Chapter) [XVI]+[\.-]?)/i
+  )
   fields = []
   vern = []
   unmatched_vern = []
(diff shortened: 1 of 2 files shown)


spec/lib/traject/config/note_fields_spec.rb
```diff
@@ -165,6 +165,23 @@
...
+              MARC::DataField.new(
+                '505', ' ', ' ',
+                MARC::Subfield.new('a', 'aaa -- Word 10 Another Word -- ccc')
+              )
+            )
+          end
+        end
+
+        it 'does not split on a number that is not intended to be the beginning of a new chapter/line' do
+          expect(result_field.first[:fields].first).to eq ['aaa', 'Word 10 Another Word', 'ccc']
```
jkeck: FWIW, before this change this would have returned 

```ruby
['aaa', 'Word', '10 Another Word', 'ccc']
```

lib/traject/config/sirsi_config.rb
```diff
@@ -1918,7 +1918,13 @@ def marc_008_date(byte6values, byte_range, u_replacement)
 #    vernacular: [['The same, but pulled from the matched vernacular fields']]
 #    unmatched_vernacular: [['The same, but pulled from any unmatched vernacular fields']]
 to_field 'toc_struct' do |marc, accumulator|
-  formatted_chapter_regex = Regexp.union(/[^\S]--[^\S]/, /      /, /(?=(?:Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i,  /(?=(?<!Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i, /(?=(?:Appendix|Section|Chapter) [XVI]+[\.-]?)/i)
+  formatted_chapter_regex = Regexp.union(
+    /[^\S]--[^\S]/,
+    /      /,
+    /(?=(?:Chapter|Section|Appendix|Part|v\.) \d+[:\.-]\s+)/i,
```
cbeer: I wonder if this was copy-pasta and should have been:

```
/(?=(?:Chapter|Section|Appendix|Part|v\.) \d+[:\.-]?\s+)/i,
/(?=(?<!Chapter|Section|Appendix|Part|v\.) \d+[:\.-]\s+)/i
```

(Punctuation required if it isn't preceded by one of those words). I'm 👍 to shipping this now, but I wonder if we should grab a ton of real-world data to throw at it too?

jkeck:
FWIW, I grabbed 6K ckeys for records that contained a ToC and pulled on the MARC 505/905s from those records and ran that through the indexer before and after this change, and then spit out the records where the ToCs were different.  In the vast majority of cases the result after the change was an improvement.  However; there were a small percentage where things got worse (or just didn't really parse as a ToC at all).

Here are some examples:
13292813
13367897
13298517
13328352
13367897
13355329
13325227
13310157
13313336

In most of these cases, the thing that doesn't get split anymore (which is no surprise) is where a number surrounded by spaces.  This is a good thing in most cases, but in the examples above we'll no longer have different bullets for each of the sections (and I'm not sure that data is structured enough to reliably do so w/o splitting a lot of things we do not intend to).

One pattern I did notice in the ToCs where we were no longer splitting on a number surrounded by a space is that it is frequently preceded by a > 1 whitespace character.  If I have a bit of time I may try to tweak the regex a bit and re-run it over these 6K ToCs and see if the numbers improve at all.

PR #400, opened 2020-01-06
Don't update rubygems or install bundler

jkeck:
(we should have the latest by default now)


jkeck:
Closing this in favor of #401 (where I've added this commit w/ a test fix)

PR #411, opened 2020-02-19
Returns values for unmatched vernacular TOC and summary structs

### LINKED ISSUE #410: Unmatched vernacular TOC and summary struct contains subfield codes
https://github.com/sul-dlss/searchworks_traject_indexer/blob/6a422d28ab8448a68cc8751672f07972f4db11ae/lib/traject/config/sirsi_config.rb#L2021-L2035
This method returns the subfield codes, such as in ckey 10412542:
```
"unmatched_vernacular":[[
  "$g 001-026.  $t 水浒传(全26册)",
  "$g 027-041.  $t 岳飞传(全15册)",
  "$g 042-046.  $t 杨家将(全5册)",
  "$g 047.  $t 西厢记",
  "$g 048.  $t 梁山伯与祝英台", <snip>]]
```

It should return only the values of the subfield codes, like the method get_marc_vernacular, so the struct looks like:
```
"unmatched_vernacular":[[
  "001-026. 水浒传(全26册)",
  "027-041.  岳飞传(全15册)",
  "042-046. 杨家将(全5册)",
  "047. 西厢记",
  "048. 梁山伯与祝英台", <snip>]]
```

shelleydoljack:
Fixes #410

### DIFF
--- a/lib/traject/config/sirsi_config.rb
+++ b/lib/traject/config/sirsi_config.rb
@@ -2022,13 +2022,16 @@ def get_unmatched_vernacular(marc,tag)
   fields = []
   if marc['880']
     marc.find_all { |f| '880' == f.tag }.each do |field|
-      text = ""
+      text = []
       unless field['6'].nil? or !field["6"].include?("-")
         if field['6'].split("-")[1].gsub("//r","") == "00" and field['6'].split("-")[0] == tag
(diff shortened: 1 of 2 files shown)


shelleydoljack:
@jkeck I just realized that maybe the way I changed the method `get_unmatched_vernacular` might not return the right data structure. For instance, when running traject with `--debug-mode` for the marc file, the `summary_struct` field gets:
```

PR #441, opened 2020-05-05
Simplify and and unify logic for when a link is considered full-text.

jkeck:
We were handling "blank" a indicator2 a bit differently between the marc_links_struct and the url_fulltext field (which is used to determine of Access = Online)

This unifies that approach and simplifies it a bit.

We are no longer trying to treat "blank" indicator2 as full-text anymore in the Online facet.  We only appeared to be in the code before in the marc_links_struct, however a nil or blank indicator2 comes through as " " (failing the `.empty?` check), so was always falling through as falsey for the full-text check anyway (the behavior we likely wanted).

According to https://www.loc.gov/marc/bibliographic/bd856.html, an indicator2 of 0 or 1 is either the resource itself or a version of the resource.  We then take those 856s and cast out any links that have labels that arouse suspicion about the full-text nature of the link (e.g. table of contents).  Any other indicators are treated as not full-text, which is the behavior of what we're showing in the UI. This should have the Online facet more accurately match where we will be displaying links to the user.

### DIFF
--- a/lib/marc_links.rb
+++ b/lib/marc_links.rb
@@ -119,19 +119,18 @@ module MarcLinks
       rescue URI::InvalidURIError
         return nil
     end
+
     def link_is_fulltext?(field)
       resource_labels = ["table of contents", "abstract", "description", "sample text"]
-      if field.indicator2 == "2"
-        return false
-      elsif field.indicator2 == "0" or field.indicator2 == "1" or field.indicator2.nil? or field.indicator2.empty?
-        resource_labels.each do |resource_label|
-          return false if "#{field['3']} #{field['z']}".downcase.include?(resource_label)
-        end
-        return true
-      else
-        # this should catch bad indicators
-        return nil
+      return false unless %w[0 1].include?(field.indicator2)
+
+      # Similar logic exists in the mapping for the url_fulltext field in sirsi traject config.
+      # They need to remain the same (or should be refactored to use the same code in the future)
+      resource_labels.each do |resource_label|
+        return false if "#{field['3']} #{field['z']}".downcase.include?(resource_label)
       end
+
+      true
     end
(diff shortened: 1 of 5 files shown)


lib/marc_links.rb
```diff
@@ -119,19 +119,16 @@ def process_link(field)
...
-          return false if "#{field['3']} #{field['z']}".downcase.include?(resource_label)
-        end
-        return true
-      else
-        # this should catch bad indicators
-        return nil
+      return false unless %w[0 1].include?(field.indicator2)
+
+      resource_labels.each do |resource_label|
+        return false if "#{field['3']} #{field['z']}".downcase.include?(resource_label)
```
jkeck, line 128: I could make this 
```suggestion
        return false if [field['3'], field['z']].any? { |v| v =~ /(table of contents|abstract|description|sample text)/i}
```
to match what we do in `sirsi_config` and make them even more similar

jkeck:
An example of where this is currently going wrong: 13293614

```json
{
  "id": "13293614",
  ...
  "marc_links_struct": [
    {
      "html": "<a title='' href='http://mary-mccoy.com/'>mary-mccoy.com</a> ",
      "text": "mary-mccoy.com",
      "href": "http://mary-mccoy.com/",
      "fulltext": null,
      ...
    }
  ],
  ...
  "access_facet": [
    "At the Library",
    "Online"
  ],
  ...
}
```

lib/marc_links.rb
mejackreed: Do you think if we updated this in one place we would do it in the other? I like having the consistency of the approach. 👍

jkeck: I'll add a comment to ensure that we do.

PR #442, opened 2020-05-08
Allow indexing to point at non-prod purl hosts

cbeer:


lib/traject/config/contentsearch_config.rb
```diff
@@ -0,0 +1,67 @@
+$LOAD_PATH << File.expand_path('../..', __dir__)
```
mejackreed, line 1: Was this file meant to be part of this PR?

cbeer: nope.

PR #444, opened 2020-05-14
Add "Online" to the access_facet when an item is publicly available...

jkeck:
...in HathiTrust.

Connected to sul-dlss/SearchWorks#2455

### DIFF
--- a/lib/traject/config/sirsi_config.rb
+++ b/lib/traject/config/sirsi_config.rb
@@ -1456,6 +1456,10 @@ to_field 'access_facet' do |record, accumulator, context|
   accumulator << 'On order' if accumulator.empty?
   accumulator << 'Online' if context.output_hash['url_fulltext']
   accumulator << 'Online' if context.output_hash['url_sfx']
+  # There is similar logic in the SearchWorks HathiTrustLinks class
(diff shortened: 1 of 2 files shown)


lib/traject/config/sirsi_config.rb
```diff
@@ -1400,6 +1456,10 @@ def marc_008_date(byte6values, byte_range, u_replacement)
   accumulator << 'On order' if accumulator.empty?
   accumulator << 'Online' if context.output_hash['url_fulltext']
   accumulator << 'Online' if context.output_hash['url_sfx']
+  # There is similar logic in the SearchWorks HathiTrustLinks class
+  if context.output_hash['ht_access_sim']&.none? { |v| v == 'deny' || %w[allow:icus allow:pdus].include?(v) }
```
cbeer, line 1460: Do we need to worry about `ht_access_sim` being empty (because `[].none? == true`)? Maybe it can't happen?

jkeck: It couldn't hurt!

cbeer: 🤷‍♂️ ; I remember it being an issue over in searchworks, but, yeah, can't hurt.

PR #458, opened 2020-06-25
Support indexing collections as individual records in EarthWorks

mejackreed:
Part of the work for https://github.com/sul-dlss/earthworks/issues/592

### DIFF
--- a/lib/traject/config/geo_config.rb
+++ b/lib/traject/config/geo_config.rb
@@ -9,7 +9,7 @@ require 'traject/writers/solr_better_json_writer'
 require 'utils'
 require 'honeybadger'
 require 'digest/md5'
-
+require 'byebug'
 class GeoAuthorities
   def self.formats
     {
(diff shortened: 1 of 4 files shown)


lib/traject/config/geo_config.rb
```diff
@@ -9,7 +9,7 @@
 require 'utils'
 require 'honeybadger'
 require 'digest/md5'
-
+require 'byebug'
```
cbeer, line 12:
```suggestion
```

spec/integration/geo_config_spec.rb
```diff
@@ -58,7 +58,7 @@ def stub_mods_request(druid, body)
     it 'contains subject metadata' do
       expect(result['dc_subject_sm']).to include('Hot springs')
     end
-    it 'contains description with abstract and notes' do
+    it 'contains description with abstract and es' do
```
cbeer, line 61:
```suggestion
    it 'contains description with abstract and notes' do
```

PR #460, opened 2020-06-26
Add a reference to the collection object for geo objects.

### LINKED ISSUE #459: GeoIndexer: Add dc_source_sm for content-type="geo" for its collection
For example, the record `mc977kq8162` should have the following field indexed as part of its geoblacklight-schema record:

```json
"dc_source_sm": ["stanford-bq589tv8583"],
```

The druid should be retrievable from the public xml. It is ok if the collection is not indexed, the application will consider this a no-op.

jkeck:
Closes #459

### DIFF
--- a/lib/traject/config/geo_config.rb
+++ b/lib/traject/config/geo_config.rb
@@ -314,6 +314,15 @@ to_field 'solr_year_i' do |record, accumulator|
   accumulator << record.stanford_mods.pub_year_int if accumulator.empty?
 end
 
+to_field 'dc_source_sm' do |record, accumulator|
+  next unless record.dor_content_type == 'geo'
+  next unless record.collections && record.collections.any?
+
+  record.collections.each do |collection|
+    accumulator << "stanford-#{collection.druid}"
+  end
+end
+
 to_field 'dct_isPartOf_sm', mods_xpath('mods:relatedItem[@type="host"]/mods:titleInfo/mods:title') do |record, accumulator|
   accumulator.flatten!.map!(&:text)
 end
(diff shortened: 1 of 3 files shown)


lib/traject/config/geo_config.rb
```diff
@@ -314,6 +320,12 @@ module Constants
   accumulator << record.stanford_mods.pub_year_int if accumulator.empty?
 end
 
+to_field 'dc_source_sm', mods_xpath('string(//rdf:RDF/rdf:Description/fedora:isMemberOfCollection/@rdf:resource)') do |record, accumulator|
```
jkeck, line 323: How do you `xpath` again?

mejackreed: Can we use this? https://github.com/sul-dlss/searchworks_traject_indexer/blob/master/lib/sdr_stuff.rb#L165-L169

jkeck: Meaning use that same xpath, or to extract that method into something that both indexers can use?

mejackreed: I _think_ we already have access to that method, no?

mejackreed: We use a related method like this: https://github.com/sul-dlss/searchworks_traject_indexer/blob/master/lib/traject/config/geo_config.rb#L156 ?

jkeck: Ah, I was miss-reading that file defined multiple classes and thought. that was tied to the SDR config.  We're not using that class anywhere, but I'll see how we've implemented it elsewhere.  👍

jkeck: And now I see that record is that thing.  It has been awhile since I have trajected, so took me a minute to remember how any of this works.

PR #489, opened 2020-10-05
Upgrade kostya-sigar to > 2.0.7 as there is an issue installing on MacOS

mejackreed:
See https://github.com/kostya/sigar/pull/9#issuecomment-703508839


jkeck:
Ugh, doesn't seem like it's happy under jruby.

PR #517, opened 2021-02-18
Use consistent (and correct) parsing for  links; fixes #412

### LINKED ISSUE #412: script identification code keeps non-Latin parallel fields from displaying
If a script identification code appears in the 880, the non-Latin parallel fields don't display properly.  Shelley did some analysis and research. She writes:

If we add `.gsub("/$1","")` to this line: https://github.com/sul-dlss/searchworks_traject_indexer/blob/6a422d28ab8448a68cc8751672f07972f4db11ae/lib/traject/config/sirsi_config.rb#L765 so that it reads:
vern_field = vern_fields.find { |f| f['6'] && f['6'].split("-")[1].gsub("//r","").gsub("/$1","") == field['6'].split("-")[1].gsub("//r","").gsub("/$1","") } if field['6']

then then author_struct solr field will get the vernacular field data:
13423466 author_struct {"contributors":[{"link":"Dong nan Ya shen xue jiao yu ji jin hui.","search":"\"Dong nan Ya shen xue jiao yu ji jin hui.\"","pre_text":"","post_text":"","authorities":[],"rwo":[],"vern":{"link":"東南亞神學教育基金會.","search":"\"東南亞神學教育基金會.\"","pre_text":"","post_text":"","authorities":[],"rwo":[]}}]}

We already remove the orientation code (/r) from this field so we could probably do the same for the script identification code. Since there are multiple script identification codes (see $6 - Linkage > Script identification code at https://www.loc.gov/marc/bibliographic/ecbdcntf.html for more), we might want to strip them all out. We should probably only add the list of codes that are designated for MARC-8 since they should not be in our records (Unicode environment).

cbeer:


jkeck (approved):
LGTM, pulling the #parse_linkage out is nice!

PR #538, opened 2021-05-14
Index pub_year_ss

cbeer:
This was apparently a good idea 5 years ago: https://github.com/sul-dlss/SearchWorks/blob/master/app/helpers/results_document_helper.rb#L12

🤞  i got the logic ported over correctly.


camillevilla (approved):
🤞

PR #636, opened 2022-07-08
Remove regex_split workaround method

marlo-longley:
Small PR to cleanup an outdated method that I came across doing work for #635
The issue that the workaround addressed in jruby has been closed.


marlo-longley:
@cbeer I removed the second workaround function as well.

PR #646, opened 2022-07-28
Add a client for Folio/OKAPI requests

cbeer:
Look to push upstream separately: 
- [x] #653: refactor 999 field parsing to consistently use the clipboarded `holdings`
- [x] https://github.com/sul-dlss/searchworks_traject_indexer/pull/650
- [x] #651 export marc JSON data in sirsi config (`marc_json_struct`)

TODO:
- [ ] #662
- [ ] should we look somewhere besides the folio bib record metadata for the 916b last cataloged date? (see https://github.com/sul-dlss/searchworks_traject_indexer/issues/662#issuecomment-1204543792)
- [x] consolidate the sirsi + folio configs
- [x] add folio-specific indexing for data from the `999`: `format_physical_ssim` (partial), `location_facet` (partial), `barcode_search`
- [x] #652 
- [ ] #664 
- [ ] handle `item_display` public note (999o) data. 
- [ ] #663 
- what 9xx fields are indexed into `all_search`; where will they come out with FOLIO? `905` `908` `920` `986` `979`
- is "materialType" the correct field for the item type in FOLIO?
- where is the 856 data?
- How do we handle + map item statuses? "at the bindery", "lost"?

Future work?
- [ ] change hashed_id_ssim implementation to use the instance uuid, since it's already a hash of some other fields?

Other questions:
- why did we subclass MARC::FastXMLWriter and is the behavior in SolrMarcStyleFastXMLWriter still required?
- can `effectiveShelvingOrder` replace our shelfkeys (and get rid of weird lopping code) ? help with shelve-by-title enumeration?
- does mod_graphql save us from having to do some value/code lookups?


cbeer:
Closing in favor of #674

PR #654, opened 2022-08-01
Convert MARC 999 direct refs in sirsi_config to use holdings class

### LINKED ISSUE #653: Use holdings instead of 999 field directly
Consider explicitly store `holdings` on the traject clipboard (currently memoized there) and use it everywhere we're working with holdings or data from the 999 fields.

Update subfield access to the 999 fields to use named accessors (e.g. `callnumber` instead of `tag['a']`

Add additional named accessors as needed.

cbeer:
Fixes #653


thatbudakguy:
i think this might need to be re-picked one more time to 54ef844bc4a5241b0006e306a5fdefa660888274 to get the change we later found to `building_facet`

corylown (approved):

PR #668, opened 2022-08-04
Use double quotes to enclose HTML attributes in marc_links_struct; fixes #2962

corylown:
The other way to fix this would be to encode the single quote as `%27`, but it seems preferable to just use double quotes to enclose the HTML attributes since un-encoded double quotes are not valid in URLs.

Before:
```
<a title='' href='https://www.africamuseum.be/sites/default/files/media/docs/research/publications/rmca/online/documents-social-sciences-humanities/L'Echo_des_Monts_et_Vallees.pdf'>www.africamuseum.be</a>
```
After:
```
<a title="" href="https://www.africamuseum.be/sites/default/files/media/docs/research/publications/rmca/online/documents-social-sciences-humanities/L'Echo_des_Monts_et_Vallees.pdf">www.africamuseum.be</a>
```

Fixes: https://github.com/sul-dlss/SearchWorks/issues/2962

### DIFF
--- a/lib/marc_links.rb
+++ b/lib/marc_links.rb
@@ -18,8 +18,11 @@ module MarcLinks
       {
         version: '0.1',
 
-        html: [%Q(<a title="#{link_title}" href="#{link_field['u']}">#{link_text}</a>), "#{'(source: Casalini)' if link_is_casalini?}", (%Q( <span class="additional-link-text">#{additional_text}</span>) if additional_text)].compact.join(' '),
-        text: [link_text, "#{'(source: Casalini)' if link_is_casalini?}", (" <span class='additional-link-text'>#{additional_text}</span>" if additional_text)].compact.join(' ').strip,
+        html: [%Q(<a title="#{link_title_html_escaped}" href="#{link_field_html_escaped}">#{link_text_html_escaped}</a>),
+               "#{'(source: Casalini)' if link_is_casalini?}",
+               (%Q( <span class="additional-link-text">#{additional_text_html_escaped}</span>) if additional_text)].compact.join(' '),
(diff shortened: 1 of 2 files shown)


lib/marc_links.rb
```diff
@@ -18,7 +18,7 @@ def as_h
       {
         version: '0.1',
 
-        html: ["<a title='#{link_title}' href='#{link_field["u"]}'>#{link_text}</a>", "#{'(source: Casalini)' if link_is_casalini?}", (" <span class='additional-link-text'>#{additional_text}</span>" if additional_text)].compact.join(' '),
+        html: [%Q(<a title="#{link_title}" href="#{link_field['u']}">#{link_text}</a>), "#{'(source: Casalini)' if link_is_casalini?}", (%Q( <span class="additional-link-text">#{additional_text}</span>) if additional_text)].compact.join(' '),
```
cbeer, line 21: Should we just go ahead and treat all those values as untrusted and HTML escape them? Or, maybe compose the link downstream in Searchworks in the first place?

thatbudakguy: What would it take to compose it in SearchWorks? that sounds like a better approach, potentially

corylown: I agree that SearchWorks would be a better place for generating the html links. The linking test coverage in SW seems pretty good, but it will definitely take some time to untangle things on the SW side.

corylown: I pushed up a change to this PR to HTML escape the strings that get embedded in these chunks of HTML. Moving the HTML link generation into SearchWorks makes a lot of sense to me, but it'll take some time a consideration about how to approach it. Will create a separate issue for that.

PR #676, opened 2022-08-05
Handle Lane mappings

cbeer:


lib/folio_record.rb
```diff
@@ -17,10 +17,61 @@ def instance_id
     record.dig('externalIdsHolder', 'instanceId')
   end
 
+  def sirsi_holdings
+    @sirsi_holdings ||= items.map do |item|
```
thatbudakguy, line 21: this is kind of weird. at this point I almost feel like we should make a `FolioHolding` that does nothing other than inherit from `SirsiHolding`, just so we stop ending up with odd things like this that we might forget to change the names of

spec/integration/compare_sirsi_and_folio_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'spec_helper'
 
 describe 'comparing against a well-known location full of documents generated by solrmarc' do
```
thatbudakguy, line 3: this describe text doesn't seem right for the test

thatbudakguy (approved):

spec/integration/compare_sirsi_and_folio_spec.rb
cbeer: I agree, but I don't think it's appropriate to change the text in this PR?

thatbudakguy: fair!

PR #677, opened 2022-08-06
Let traject load the sirsi config, 

cbeer:
…and do it after we set up our settings so we don't have to jump through hoops


thatbudakguy (approved):
👍🏻

Issue #662, opened 2022-08-03
Preprocess FOLIO bib + item data to remove staff information


cbeer:
Note: It's unclear how much of the symphony cleanup is applicable to FOLIO.

Original source:
https://consul.stanford.edu/display/SYSTEMS/Data+Feed+to+SearchWorks

- [x] remove "junk tags" <details><summary>(full list)</summary>
001
003
004
005
852
853
863
866
867
868
901
910
918
923
924
930
935
940
948
949
950
955
960
962
980
981
983
990
993
998
999
</details>

- [x] validate whether we use the 916 tag directly in searchworks (not present in FOLIO, using separate mapping for date_cataloged)
- [x] Price stripped for item records (was in 999 $p; unknown FOLIO equivalent, if any)
- [x] Date/time of export: Date added to 005 field (validate whether it is use for anything)

thatbudakguy:
it sure doesn't look like we're using the 005 field. i can't find any traject fields defined for it, and while our MARC fixture data has it, there doesn't appear to be anywhere referencing that data. if we later decided we needed it, we could maybe do something clever with the `last_response_date` that the `FolioClient` knows.

thatbudakguy:
re: price i think this is unrelated in FOLIO. the only place that information seems like it's available is through the [PO Line API](https://docs.folio.org/docs/acquisitions/orders/#cost-details), and I don't even know how or whether item records are linked to the POs that acquired them.

thatbudakguy:
re: `date_cataloged`, we [use it in SW](https://github.com/sul-dlss/SearchWorks/blob/6bb2822c835203fb436208b6e8e0c2dd03eae803/app/controllers/catalog_controller.rb#L453) to power sorting by that value, which looks like this:


![Image](https://user-images.githubusercontent.com/4924494/182722348-4265dd82-6a51-4a2e-8a30-a6a9f60fd021.png)

it looks like values in that field are already iso8601 timestamps like `2022-08-02T00:00:00Z`, so I think we're good there. i don't think we can control this value in FOLIO though, so it might lead to a little bit of weirdness after the migration, when everything in the catalog effectively has the same (or close to the same) "catalog date" (i.e. the point at which the record was migrated into FOLIO).

cbeer:
> it might lead to a little bit of weirdness after the migration, when everything in the catalog effectively has the same (or close to the same) "catalog date"

[...]

PR #684, opened 2022-08-15
Strip junk tags from folio records

thatbudakguy:
See #662 for a list of junk tags, and https://consul.stanford.edu/display/SYSTEMS/Data+Feed+to+SearchWorks#DataFeedtoSearchWorks-PostProcessing for a detailed explanation of processing steps that we're hoping to move to traject (of which this is one).

- Strip parsedRecord from folio_json_struct
- Implement junk tag stripping for Folio records

We ultimately decided not to strip some of the things that were in the list, including 001, 003, and the 8xx fields. It turned out that some of this data was actually used (e.g. for MHLDs in the case of the 8xx) or at least not harmful (in the case of 001 and 003).

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -171,15 +171,3 @@ end
 to_field 'holdings_json_struct' do |record, accumulator|
   accumulator << JSON.generate(record.items_and_holdings) if record.items_and_holdings
 end
-
-
-## Postprocessing
-# https://consul.stanford.edu/display/SYSTEMS/Data+Feed+to+SearchWorks
-
-# Remove "junk tags"
-each_record do |record, context|
-end
-
-# Add date accessed (from client) to marc 005
-each_record do |record, context|
-end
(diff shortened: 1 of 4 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -165,9 +165,21 @@ def holdings(record, context)
...
+## Postprocessing
+# https://consul.stanford.edu/display/SYSTEMS/Data+Feed+to+SearchWorks
+
+# Remove "junk tags"
+each_record do |record, context|
+end
+
+# Add date accessed (from client) to marc 005
+each_record do |record, context|
+end
```
corylown, line 185: I'm guessing leaving this here is intended as a placeholder for future work?

corylown (approved):

lib/traject/config/folio_config.rb
cbeer, line 185:
```suggestion
```

Looks place-holdery to me.

thatbudakguy: yep, and no longer relevant. removed it, thx

cbeer (approved):

PR #692, opened 2022-08-19
Query the FOLIO postgres database directly to collect data

cbeer:
We get a very significant (15x+++) performance boost going directly to the database instead of querying FOLIO APIs. We've carefully constructed the queries (by cribbing from FOLIO's stored functions) to generally align well with the FOLIO API output so they are relatively interchangeable in terms of the data they return in case we need to bail on this low-level access some day.


corylown (approved):
Seems good? That query though -- I had to look up the `#>>` notation 😹

PR #698, opened 2022-08-23
Wire up FOLIO date_cataloged to instance catalogedDate

### LINKED ISSUE #691: Populate date_cataloged with correct value for FOLIO records
see https://github.com/sul-dlss/searchworks_traject_indexer/issues/662#issuecomment-1204543792 and discussion for context.

we need to use the `catalogedDate` field on FOLIO instance records to populate `date_cataloged`, rather than the metadata block on the source record. this will help avoid the potential issue where everything migrated from Symphony has a catalog date of the migration time, rather than its actual first catalog date.

thatbudakguy:
Fixes #691

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -146,21 +146,11 @@ to_field 'barcode_search' do |record, accumulator, context|
   end
 end
 
-# TODO: is this the right place to get this from? use instanceRecord or holdingsRecord instead?
-# when are these timestamps updated in FOLIO?
+# guard against dates with 'u' coming out of MARC 008 fields
 to_field 'date_cataloged' do |record, accumulator|
-  timestamp = record.record.dig('metadata', 'createdDate')
-
-  # when requesting via /source-storage/stream/source-records, you get a unix epoch (e.g. 1658265346880)
-  # but when requesting via /source-storage/source-records, you get an ISO8601 string (e.g. "2020-01-01T00:00:00Z")
-  # we also need to always convert it to UTC for solr
-  if timestamp.is_a?(Numeric)
-    accumulator << Time.at(timestamp).utc.iso8601
-  elsif timestamp
-    accumulator << Time.iso8601(timestamp).utc.iso8601
-  end
+  timestamp = record.instance['catalogedDate']
+  accumulator << Date.parse(timestamp).to_time.utc.at_beginning_of_day.iso8601 if timestamp && !timestamp.include?('u')
 end
-
 
 # add folio to the collection list; searchworks has some dependencies on this value,
 # so for now, we're just appending 'folio' to the list.
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -149,7 +149,7 @@ end
 # guard against dates with 'u' coming out of MARC 008 fields
 to_field 'date_cataloged' do |record, accumulator|
   timestamp = record.instance['catalogedDate']
-  accumulator << Date.parse(timestamp).to_time.utc.at_beginning_of_day.iso8601 if timestamp && !timestamp.include?('u')
+  accumulator << Time.parse(timestamp).utc.at_beginning_of_day.iso8601 if timestamp =~ /\d{4}-\d{2}-\d{2}/
 end
 
 # add folio to the collection list; searchworks has some dependencies on this value,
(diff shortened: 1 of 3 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -146,18 +146,18 @@ def holdings(record, context)
...
   # we also need to always convert it to UTC for solr
-  if timestamp.is_a?(Numeric)
-    accumulator << Time.at(timestamp).utc.iso8601
-  elsif timestamp
-    accumulator << Time.iso8601(timestamp).utc.iso8601
+  # division by 1000 is to remove milliseconds from the epoch so it parses correctly
+  case timestamp
+  when Numeric
+    accumulator << Time.at(timestamp / 1000).utc.iso8601
+  when String && (timestamp !~ /NEVER/i)
```
cbeer, line 159: Are we populating `catalogedDate` with `NEVER` in FOLIO, or is that a symphony artifact?

thatbudakguy: in slack, @shelleydoljack said:
> coming out of symphony, it’ll be 0. should we leave it that way? i think we’re converting that to NEVER in our export.

so i think we could do it either way — i opted for this for parity with the sirsi config, but maybe there's something simpler.

thatbudakguy: i think if we don't replace 0 with `NEVER`, the current logic will put it in the `Numeric` bucket and it'll become the 1970 date, which seems less accurate than just not indexing a value there.

lib/traject/config/folio_config.rb
```diff
@@ -146,18 +146,18 @@ def holdings(record, context)
   end
 end
 
-# TODO: is this the right place to get this from? use instanceRecord or holdingsRecord instead?
-# when are these timestamps updated in FOLIO?
 to_field 'date_cataloged' do |record, accumulator|
-  timestamp = record.record.dig('metadata', 'createdDate')
+  timestamp = record.instance['catalogedDate'] || record.instance.dig('metadata', 'createdDate')
```
shelleydoljack, line 150: It looks like when there is no cataloged data in FOLIO, the field is not returned in the response for something like `GET https://${OKAPI_URL}/inventory/instances?query=hrid=="in00000000224"`. I don't think we should use createdDate if `record.instance['catalogedDate']` is missing because it could put way too many things at the top of the sort when we batch-load records into the system, making "new to the library" sort of useless.

lib/traject/config/folio_config.rb
shelleydoljack: Coming out of Symphony, when date cataloged is never, the API returns 0. When migrating to FOLIO, this should translate to what FOLIO does, which is not write a value to the instance field catalogedDate. So, for FOLIO, line 150 and 158 matter and Symphony-migrated records in FOLIO won't need line 159. However, because IndexData is taking date cataloged from MARC 008, which could be a string (e.g. "19uu-uu-uu", but that's a migration error, they should not write to that field when data looks like this), then we should account for that in some way, like not adding it to the `date_cataloged` solr field.

thatbudakguy:
fixed this up after pairing with @shelleydoljack. we no longer try to populate the field if `catalogedDate` is empty because it could cause weird behavior for the date facet, and we guard against dates with `u` in them coming out of MARC, which we saw happening for Lane records that IndexData handled.

also added tests.

lib/traject/config/folio_config.rb
```diff
@@ -146,22 +146,12 @@ def holdings(record, context)
...
-  # when requesting via /source-storage/stream/source-records, you get a unix epoch (e.g. 1658265346880)
-  # but when requesting via /source-storage/source-records, you get an ISO8601 string (e.g. "2020-01-01T00:00:00Z")
-  # we also need to always convert it to UTC for solr
-  if timestamp.is_a?(Numeric)
-    accumulator << Time.at(timestamp).utc.iso8601
-  elsif timestamp
-    accumulator << Time.iso8601(timestamp).utc.iso8601
-  end
+  timestamp = record.instance['catalogedDate']
+  accumulator << Date.parse(timestamp).to_time.utc.at_beginning_of_day.iso8601 if timestamp && !timestamp.include?('u')
```
cbeer, line 152: Maybe it's better to check for what we want instead of playing whack-a-mole with the stuff we don't?

```suggestion
  accumulator << Time.parse(timestamp).utc.at_beginning_of_day.iso8601 if timestamp =~ /\d{4}-\d{2}-\d{2}/
```

PR #705, opened 2022-08-24
Parallelize the postgres data load into kafka by sharding on the instance id

cbeer:
Parallelizing with 8 threads is ~4x faster; it's unclear at the moment why it isn't even better, but hopefully this lets us fail faster when doing data loading.


thatbudakguy (approved):
good call on all the `ensure`

PR #710, opened 2022-08-26
Trim metadata from some objects

cbeer:


thatbudakguy (approved):
does the ` - ` operator actually subtract keys from the object? that's wild.

cbeer:
Yup: https://www.postgresql.org/docs/13/functions-json.html#FUNCTIONS-JSONB-OP-TABLE

Issue #681, opened 2022-08-11
Index records that only have instance data (no SRS data)


cbeer:
- [ ] course reserves data that "catalog" an instructor copy
- [ ] on-order acq without MARC records
- [ ] equipment?

Is there a way to get MARC data for this, or do we have to deal with inventory/instance JSON data

thatbudakguy:
@ahafele will create an on-order acq without MARC records and an equipment record for testing.

saseestone:
Getting an error today when trying to add any Fast Add item (from Courses app or Inventory app). Will report in folio-support channel too.

ahafele:
I'm unable to create "on-order acq without MARC records and an equipment record" at the moment because Inventory is not working correctly. I will try again once those issues are resolved.

PR #716, opened 2022-09-13
Index records without holdings, items, or source record data

cbeer:
Proof of concept for #681. Will need an actual spec for the instance-derived MARC records some day.


thatbudakguy (approved):
lgtm as a first pass!

PR #718, opened 2022-09-21
Add ItemDisplay class to contain logic to create item_display values; related cleanup.

corylown:
Still working on some tests...


lib/traject/config/sirsi_config.rb
```diff
@@ -2219,73 +2138,22 @@ def holdings(record, context)
...
-    else
-      scheme = ''
-      shelfkey = ''
-      volume_sort = ''
-      reverse_shelfkey = ''
-      lopped_call_number = holding.call_number.to_s
-    end
-
-    current_location = holding.current_location
-    current_location = 'ON-ORDER' if holding.is_on_order? && holding.current_location && !holding.current_location.empty? && holding.home_location != 'ON-ORDER' && holding.home_location != 'INPROCESS'
```
corylown, line 2274: I'm pretty sure this will _never_ get set because `holding.is_on_order?` only evaluates to `true` if `holding.current_location` is _already_ set to `ON-ORDER` OR if `holding.home_location` is 'ON-ORDER' in which case the latter `holding.home_location != 'ON-ORDER'` condition will not pass. 😕

PR #734, opened 2022-12-12
Language map

arcadiafalcone:
Updates language code mapping to match the current MARC languages standard. Deprecated codes are marked for future reference, but cannot be removed until legacy data has been remediated.

### DIFF
--- a/lib/translation_maps/marc_languages.properties
+++ b/lib/translation_maps/marc_languages.properties
@@ -11,6 +11,7 @@ afa = Afroasiatic (Other)
 afh = Afrihili (Artificial language)
 afr = Afrikaans
 ain = Ainu
+# ajm is deprecated
 ajm = Aljamía
 aka = Akan
 akk = Akkadian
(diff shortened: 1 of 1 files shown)


cbeer:
Do we need to coordinate this with a similar language map change for MODS records?

arcadiafalcone:
MARC and MODS are already out of sync so it doesn't need to happen at the same time, but similar updates will need to be made on the MODS side as well (I'm working on it).

Is there automatic reindexing that will propagate the changes?

lib/translation_maps/marc_languages.properties
```diff
@@ -300,19 +318,21 @@ men = Mende
...
 mni = Manipuri
 mno = Manobo languages
 moh = Mohawk
+# mol is deprecated
 mol = Moldavian
 mon = Mongolian
-mos = Moore
-#mul = Multiple languages
+mos = Mooré
+mul = Multiple languages
```
cbeer, line 335: There are some tests that suggest we explicitly wanted to hide `mul` / `Multiple languages` (and `mis`, presumably). Have we decided to re-enable it, or was this a mistake?

arcadiafalcone:
They are enabled in the currently implemented mapping, so I restored them here as well.

Issue #765, opened 2023-03-29
Addition to LC classification code translation table


cbeer:
Per Vitus:

The Law Library is requesting that the following K subclass be added to the LC classification code translation table used in SearchWorks' Call Number facet:

 KVL = KVL - Law of Cook Islands, Easter Island

SW-3761

PR #779, opened 2023-04-13
Adding KVL to LC classification code translation table

hudajkhan:


hudajkhan:
I tried this change out by using the searchworks traject indexer to index a MARC file that originally had a KVN classification which I changed to KVL.  The resulting Solr document had KVL in the call number facet (callnum_facet_hsim), callnum_search, and c_assigned_callnum_ssim fields (and represented wherever the call number is included).  A local version of SearchWorks also showed the call number facet values with KVL.

hudajkhan:
Connected to #765

PR #780, opened 2023-04-14
Change contents label according to 505 indicator

marlo-longley:
Fixes https://github.com/sul-dlss/SearchWorks/issues/3103

See https://www.loc.gov/marc/bibliographic/bd505.html for indicator/status names

Screenshot 2023-04-14 at 10 51 22

### DIFF
--- a/lib/traject/config/sirsi_config.rb
+++ b/lib/traject/config/sirsi_config.rb
@@ -1555,6 +1555,7 @@ to_field 'toc_struct' do |marc, accumulator|
   fields = []
   vern = []
   unmatched_vern = []
+  label = 'Contents'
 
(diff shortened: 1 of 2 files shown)


cbeer:
I wonder if we also want to map indicator1=1 to "Partial contents".

I mean, I don't know what "Partial contents" means either, but the description of "Incomplete contents" seems pretty library-workflow-centric, doesn't it?

@saseestone ?

marlo-longley:
I rebased to master and the new rubocop rules, and changed "Incomplete" to "Partial" contents in the label.

Jenkins failed with  
```
ERROR: Error fetching remote repo 'origin'
hudson.plugins.git.GitException: Failed to fetch from https://github.com/SUL-DLSS/searchworks_traject_indexer.git
```

PR #789, opened 2023-04-18
Remove unused direct dependency rexml

jcoyne:
It is a transitive dependency, so we don't need it in Gemfile


corylown:
Needs rebase

PR #803, opened 2023-04-20
Use rspec idioms for conditional tests

jcoyne:


jcoyne:
@cbeer sure, I didn't notice that.

PR #809, opened 2023-04-21
Add a procfile for background jobs

### LINKED ISSUE #643: Convert eye-managed processes to use foreman instead
`eye` is giving us some grief deploying under ruby 3.x (hanging processes, unable to find `eye` binary, etc). Now that we're in a systemd environment, maybe user-level systemd scripts are the way to go?

See #632 for an example of the weird kind of errors that come up.

jcoyne:
Fixes #643


corylown (approved):

README.md
```diff
@@ -73,22 +95,7 @@ processes:
...
-    earthworks-stage-indexer_0 .... up  (12:07, 0%, 73Mb, <1218016>)
-    folio_dev_indexer_0 ........... up  (14:34, 6%, 137Mb, <1386323>)
-    marc_bodoni_dev_indexer_0 ..... up  (12:07, 0%, 2731Mb, <1223573>)
-    marc_morison_dev_indexer_0 .... up  (12:07, 0%, 1253Mb, <1223963>)
-    sw_dev_indexer_0 .............. up  (12:07, 0%, 75Mb, <1224054>)
-    sw_dev_indexer_1 .............. up  (12:07, 0%, 73Mb, <1224303>)
-    sw_preview_stage_indexer_0 .... up  (12:07, 0%, 73Mb, <1224638>)
-    sw_preview_stage_indexer_1 .... up  (12:07, 0%, 73Mb, <1224861>)
-```
-you can stop and start daemons with e.g. `eye stop sw_dev_indexer`. note that it may take some time for all the processes to start and stop. for more information on eye, use `eye help` or see the [eye wiki](https://github.com/kostya/eye/wiki).
```
thatbudakguy, line 91: could we replace this with the equivalent systemd command to check service status? I imagine it's `sudo systemctl status traject` or something like that, but would be good to document. also copying the `systemctl start`/`stop` commands here would be nice

README.md
```diff
@@ -29,11 +29,33 @@ bundle exec rake
...
+```
+ksu
+journalctl -u traject-marc_bodoni_dev_indexer.1.service -e
+```
+
 ## indexing data
 indexing is a multi-step process:
 1. an extractor process publishes data to be indexed to a [kafka](https://kafka.apache.org/) topic
-2. a daemon run by [eye](https://github.com/kostya/eye) consumes data from the kafka topic and invokes traject
+2. Systemd runs the various traject services
```
thatbudakguy, line 56: don't the services still do the same thing? i think it's useful to note that they are consuming data from kafka to send to traject

README.md
jcoyne: I don't think they are really equivalent.  `systemctl status traject.target` is only going to show:
```
sudo systemctl status traject.target
● traject.target
     Loaded: loaded (/lib/systemd/system/traject.target; enabled; vendor preset: enabled)
     Active: active since Mon 2023-05-08 14:02:26 PDT; 1 day 19h ago

May 08 14:02:26 sw-indexing-stage-a.stanford.edu systemd[1]: Reached target traject.target.
```


that isn't going to give any of the information that eye provided.

README.md
jcoyne: It its traject itself that is acting as a kafka consumer.    But the traject service it can have any input  you configure it to have.  Systemd knows nothing about kafka, so I didn't think that belongs here.

README.md
thatbudakguy: It seems like `list-dependencies` will do this (sort of):
```
sudo systemctl list-dependencies traject.target
traject.target
● ├─traject-earthworks_stage_indexer.1.service
● ├─traject-folio_dev_indexer.1.service
● ├─traject-folio_dev_indexer.2.service
● ├─traject-folio_dev_indexer.3.service
● ├─traject-folio_dev_indexer.4.service
● ├─traject-folio_dev_indexer.5.service
● ├─traject-folio_dev_indexer.6.service
● ├─traject-folio_dev_indexer.7.service
● ├─traject-folio_dev_indexer.8.service
● ├─traject-marc_bodoni_dev_indexer.1.service
● ├─traject-marc_morison_dev_indexer.1.service
● ├─traject-sw_dev_indexer.1.service
● ├─traject-sw_dev_indexer.2.service
● ├─traject-sw_preview_stage_indexer.1.service
● └─traject-sw_preview_stage_indexer.2.service
```
at least you get the green dot for the ones that are running, which seems useful enough to document.

jcoyne: I don't have sudo access to that. How do you?
```
sudo systemctl list-dependencies traject.target
[sudo] password for indexer:
```

jcoyne: Ah, well, then the sudo isn't going to do anything. Anyways, I made a PR for that to puppet.  I didn't know about `list-dependencies` so thanks for brining that to my attention.

thatbudakguy (approved):

PR #816, opened 2023-04-27
Create a Traject::FolioJsonReader for working with local JSON files

thatbudakguy:
traject-plus offers a simple JsonReader, so this just subclasses it
to generate `FolioRecord` from the files passed to it.

Adds documentation to the README about how you could use it, after
e.g. using #823 (or the APIs) to get the record and save it as JSON.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,11 @@ another option is to use the `JsonWriter` to pipe output directly to somewhere e
 ```sh
 SOLR_URL=http://localhost:8983/solr/core-name bundle exec traject -c lib/traject/config/config_name.rb -w Traject::JsonWriter my_marc_file.marc | tail -n +2 | jq '.pub_country'
 ```
+when working with non-MARC data held locally (e.g. JSON exports from FOLIO), you can use the `FolioJsonReader` to pipe output into traject from stdin:
+```sh
+# after exporting OKAPI_URL, OKAPI_USER, OKAPI_PASSWORD
+cat record.json | bundle exec traject -c lib/traject/config/folio_config.rb -s reader_class_name=Traject::FolioJsonReader --stdin --debug-mode
+```
 
 ## environments
 ### Symphony ILS (Sirsi)
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ SOLR_URL=http://localhost:8983/solr/core-name bundle exec traject -c lib/traject
 ```
 when working with non-MARC data held locally (e.g. JSON exports from FOLIO), you can use the `FolioJsonReader` to pipe output into traject from stdin:
 ```sh
-# after exporting OKAPI_URL, OKAPI_USER, OKAPI_PASSWORD
 cat record.json | bundle exec traject -c lib/traject/config/folio_config.rb -s reader_class_name=Traject::FolioJsonReader --stdin --debug-mode
 ```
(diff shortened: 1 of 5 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -32,6 +32,7 @@
     require 'traject/readers/folio_postgres_reader'
     provide 'reader_class_name', 'Traject::FolioPostgresReader'
   else
+    require 'traject/readers/folio_json_reader'
```
thatbudakguy, line 35: it might be dumb to eagerly require this, but I wasn't sure how to detect that traject was running with `--stdin`

lib/traject/readers/folio_json_reader.rb
```diff
@@ -0,0 +1,24 @@
...
+require_relative '../../folio_record'
+
+module Traject
+  # A Traject reader for processing MARC JSON from FOLIO via stdin.
+  class FolioJsonReader < TrajectPlus::JsonReader
+    def each(&)
+      return to_enum(:each) unless block_given?
+
+      super do |record|
+        yield FolioRecord.new(record, client)
```
jcoyne, line 14: Can we make the client nil? We don't ever want this to hit the API.

thatbudakguy: done

README.md
```diff
@@ -118,6 +118,11 @@ another option is to use the `JsonWriter` to pipe output directly to somewhere e
 ```sh
 SOLR_URL=http://localhost:8983/solr/core-name bundle exec traject -c lib/traject/config/config_name.rb -w Traject::JsonWriter my_marc_file.marc | tail -n +2 | jq '.pub_country'
 ```
+when working with non-MARC data held locally (e.g. JSON exports from FOLIO), you can use the `FolioJsonReader` to pipe output into traject from stdin:
+```sh
+# after exporting OKAPI_URL, OKAPI_USER, OKAPI_PASSWORD
```
jcoyne, line 123: Is this necessary? We aren't using the API, right?

jcoyne (changes requested):

README.md
marlo-longley: We never hit the API but I believe since folio_config constructs a client behind the scenes that's why we need it.
https://github.com/sul-dlss/searchworks_traject_indexer/blob/cc8768eb71e4bb586b67a3fa7664846559caddca/lib/traject/config/folio_config.rb#L36
When I run this command without providing those ENV vars, I get:

`Error loading configuration file lib/traject/config/folio_config.rb:37 URI::InvalidURIError:bad URI(is not URI?): nil`

jcoyne: Perhaps we need another clause in the if statement there, so we don't build a client unless it is going to be used.

thatbudakguy: Reworked the statement so that we don't try to build a client if we specified `FolioJsonReader` as the reader.

lib/traject/config/folio_config.rb
thatbudakguy: fixed by [2cf1178](https://github.com/sul-dlss/searchworks_traject_indexer/pull/816/commits/2cf1178f4c1ee02005d19d6ba29b868ccddceaf8)

jcoyne (approved):

PR #835, opened 2023-05-02
Use the SolrBetterJsonWriter for folio_config

### LINKED ISSUE #663: Handle FOLIO deletes in the indexer
When a record is deleted from FOLIO (or marked as suppressed from discovery? or becomes otherwise invalid?), the record should be removed from Solr.

jcoyne:
This causes deletes to be recorded in Solr.
Fixes #663

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -23,8 +23,12 @@ require 'utils'
 
 I18n.available_locales = [:en]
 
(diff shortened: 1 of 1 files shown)


cbeer:
I'm surprised we don't get it for free from sirsi_config?

https://github.com/sul-dlss/searchworks_traject_indexer/pull/835/files#diff-93251714c685d86f273f6f69bd7e74da3770b4c810dd1fc5f1a0720033586578R55

jcoyne:
@cbeer is right, we are getting that from the sirsi config.

PR #846, opened 2023-05-03
Add a script for downloading a single FOLIO record

thatbudakguy:


marlo-longley (approved):

marlo-longley:
Nice! Handier than directly calling the `find_by_catkey`. Thanks NIck.

README.md
```diff
@@ -144,3 +144,11 @@ Data is read directly from the postgres database underlying FOLIO, using a custo
 require 'traject/readers/folio_postgres_reader'
 Traject::FolioPostgresReader.find_by_catkey('a123456, 'postgres.url' => 'postgres://[user]:[password]@localhost/okapi')
 ```
+There is also a helper script for fetching single records at a time. You need to set the `DATABASE_URL` environment variable:
+```sh
+export DATABASE_URL=postgres://[user]:[password]@localhost/okapi
+```
+...and then run the script with a catkey as an argument, optionally redirecting the output to a file:
+```sh
+./scripts/download_folio_record.rb a123456 > record.json
```
marlo-longley, line 153: Actually hold up -- we have scripts plural here. And singular in the directory.

marlo-longley (changes requested):

README.md
thatbudakguy: oooop good catch!

PR #853, opened 2023-05-05
Index information about a holding's Bound-with parent item

marlo-longley:
Even if we are not consuming this presently, we would now have it available to SearchWorks if we need to move away from relying on reading 590 MARC.

Unsure if the multiple client calls in the `bound_with_parents` method is too inefficient, open to thoughts, but via APIs it is necessary. 

Fixes https://github.com/sul-dlss/searchworks_traject_indexer/issues/881

### DIFF
--- a/lib/folio_record.rb
+++ b/lib/folio_record.rb
@@ -33,16 +33,38 @@ class FolioRecord
     @client = client
   end
 
-  def marc_record
-    @marc_record ||= MARC::Record.new_from_hash(stripped_marc_json || instance_derived_marc_record)
-  end
-
   def instance_id
     instance['id']
   end
 
   def hrid
     instance['hrid']
+  end
+
+  def marc_record
+    @marc_record ||= begin
+      record ||= MARC::Record.new_from_hash(stripped_marc_json || instance_derived_marc_record)
+
+      # if 590 with Bound-with related subfields are present, return the record as is
+      field590 = record.find { |f| f.tag == '590' }
+      subfield_a = field590&.find { |sf| sf.code == 'a' }
+      subfield_c = field590&.find { |sf| sf.code == 'c' }
+      unless field590 && subfield_a && subfield_c
+        # if 590 or one of its Bound-with related subfields is missing, and FOLIO says this record is Bound-with, append the relevant data from FOLIO
+        parents ||= bound_with_parents
+        # if Bound-with parents are found, edit the marc record
+        if parents&.any?
+          # append a new 590 and/or its subfields if not present
+          parents.each do |parent|
+            field590 = MARC::DataField.new('590', ' ', ' ') if field590.nil?
+            field590.subfields << MARC::Subfield.new('a', "#{parent['childHoldingCallNumber']} bound with #{parent['parentInstanceTitle']}") if subfield_a.nil?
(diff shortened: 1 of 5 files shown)


spec/lib/folio_record_spec.rb
```diff
@@ -33,4 +33,35 @@
...
+                                                                                                                                                                  'boundWithParts' =>
+                                                                                                                                                                    [{
+                                                                                                                                                                      'itemId' => 'd1eece03-e4b6-5bd3-b6be-3d76ae8cf96d'
+                                                                                                                                                                    }]
+                                                                                                                                                                })
+      allow(client).to receive(:get_json).with('/inventory/items/d1eece03-e4b6-5bd3-b6be-3d76ae8cf96d').and_return({
+                                                                                                                     'callNumber' => '064.8 .D191H',
```
marlo-longley, line 52: Rubocop is making me put these spaces in between lines...

thatbudakguy (commented):
I think we need to figure out the database-side piece of this so we can get it in this PR alongside the API-based version. I'm down to pair on that next week!

lib/folio_record.rb
```diff
@@ -45,6 +41,32 @@ def hrid
...
+
+      # if 590 with Bound-with related subfields are present, return the record as is
+      field590 = record.find { |f| f.tag == '590' }
+      subfield_a = field590&.find { |sf| sf.code == 'a' }
+      subfield_c = field590&.find { |sf| sf.code == 'c' }
+      unless field590 && subfield_a && subfield_c
+        # if 590 or one of its Bound-with related subfields is missing, and FOLIO says this record is Bound-with, append the relevant data from FOLIO
+        parents ||= bound_with_parents
+        # if Bound-with parents are found, edit the marc record
+        if parents && !parents.empty?
```
cbeer, line 56: Is something like this equivalent (and easier to read?)?

```suggestion
        if parents&.any?
```

thatbudakguy (approved):

marlo-longley:
@thatbudakguy I just resolved the merge conflict. Thanks for the approval.

lib/folio_record.rb
```diff
@@ -136,6 +163,28 @@ def items_and_holdings
...
+          bind = client.get_json("/inventory-storage/bound-with-parts?query=(holdingsRecordId==#{holding['id']})")
+          bind['boundWithParts'].map do |part|
+            parent = client.get_json("/inventory/items/#{part['itemId']}")
+            {
+              'childHoldingCallNumber' => holding['callNumber'],
+              'parentItemId' => parent['id'],
+              'parentItemBarcode' => parent['barcode'],
+              'parentInstanceTitle' => parent['title'],
+              # extract the instance id/ckey from the hrid to use later for displayed links
+              'parentInstanceHrid' => parent['hrid'].match(/\d+/)[0]
```
cbeer, line 180: For posterity, this won't work with newly created data in FOLIO or stuff from Lane (where the instance HRID will properly start with an L).. but maybe we cross that bridge if/when FOLIOs APIs are responsive enough that we can even consider using their APIs.

marlo-longley: @cbeer I am thinking I might just remove this whole part that does API calls until we are ready to use it.

cbeer: I think it's fine to keep it -- if nothing else, it's useful for comparing FOLIO's APIs to our direct database access

Issue #667, opened 2022-08-03
Index FOLIO course reserves data


cbeer:
We punted on all the CREZ data indexing for FOLIO. Unclear where we're getting that reserves data in FOLIO, but there's probably work for us to do there.

thatbudakguy:
@saseestone is going to create some fixture data for us in FOLIO and make sure it's associated with items for testing.

saseestone:
1 course created

Course code: AFRICAAM-105-01
Course name: Black Matters: Introduction to Black Studies

Cross-listed with course code: AFRICAAM-269-01 

It's got two items associated with it:
item barcode: 36105021019562
item barcode: 36105021837179

saseestone:
Another fixture course:

Course code: AFRICAAM-158-01
Course name: Rebellion: Black Resistance in the Caribbean
Item ID associated with course: 36105212524685

cbeer:
Thanks. We're able to pull back data from the course reserves API.

We're unsure how FOLIO handles the current location + loan period when an item is actively on reserve (or if we continue to need our Symphony data rewriting hacks)

saseestone:
Ok one fun fact - if an item is associated with a course, it's "on reserve". FOLIO doesn't have any way (currently) to prepare Item records ahead of the quarter or stage them in some way.  Unlike Symphony, there aren't reserve "status" like "prepare" or "dormant".  

So a course record has either:

A. Items associated with the course. and that therefore all the items are "on reserve"
OR
B. No items associated with the course, and therefore no items are on reserve (and we shouldn't display anything about the Course in SearchWorks.

saseestone:
@cbeer I think a step was missed in the "How to create Course records" demo. (Or I just missed seeing it.) 

It turns out that staff have to assign a temporary loan period AFTER they have associated the item with the Course.  

When I associated the first few items with courses, I did NOT update the item to use a temporary loan period.  So when I tried to checkout an item (36105021837179 - Title: Lions, black skins...) it applied the one year loan. That made me seriously scratch my head about what was going on.

[...]

Issue #766, opened 2023-03-29
Reindex records based on changes in non-instance data


cbeer:
When data changes in FOLIO that effects the sesarchworks display, we want to reindex records to make sure data is accurated displayed. The most important changes are in the bibliographic metadata or location, and we previously used the updated times from:
- [x] `sul_mod_inventory_storage.holdings_record`
- [x] `sul_mod_inventory_storage.instance`
- [x] `sul_mod_inventory_storage.item` 
- [x] `sul_mod_source_record_storage.records_lb`

There is data from other tables that may receive impactful updates:  
- [x] `sul_mod_courses.coursereserves_reserves` provides start + end dates for reserves, connects an item to a course (72 rows)
- [x] `sul_mod_courses.coursereserves_courses` provides course name + number (48 rows)
- [x] `sul_mod_courses.coursereserves_courselistings` provides instructor info (42 rows)
- [ ] `sul_mod_orders_storage.pieces` provides MHLD info (23913 rows, there are no update times in this table)
 
There's also an opportunity to use labels from various tables instead of maintaining a code->user label mapping in Searchworks:
- [ ] `sul_mod_inventory_storage.location`  
- [ ] `sul_mod_inventory_storage.loclibrary`
- [ ] `sul_mod_inventory_storage.loan_type`

thatbudakguy:
the current indexing implementation merges information from a ton of different tables, but it only looks at the `updatedDate` for instance/holding/item to decide if the record was changed.

we need to:

- [ ] determine what other fields changing constitutes a change (see above — reserve status, location name, etc). what would matter to a SearchWorks user when it gets changed?
- [ ] figure out if FOLIO is already doing any magic to bump the `updatedDate` when these things change?
- [ ] if not, ensure we can query the database for changes to those fields
- [ ] ...and if we can't query in the way we want because of the way the database is set up (likely), figure out what kind of constraints we need to enforce (e.g. you aren't allowed to rename locations and must make a new location)

thatbudakguy:
possibly worth investigating if we can use some internal message queue parts of FOLIO to listen for changes instead of/in combination with the database

cbeer:
coursereserves `updatedDate` is not an indexed field.

thatbudakguy:
big step on this one is the first checkbox above (determining how many more things we need to add to the query). if the answer is "not too many and their tables are small", possible to add our needed indices to FOLIO's database

shelleydoljack:
[...]

PR #854, opened 2023-05-05
Index course reserve data from FOLIO

thatbudakguy:
- Rename sirsi course reserves spec to distinguish it from FOLIO
- Index course reserve fields from FOLIO records

See #667, in particular the final comment, for an explanation of why we aren't modifying `item_display` and `building_facet` in FOLIO with crez info. If these assumptions are correct, then this finishes #667.

#833 should be merged first, otherwise the impact from this won't be visible because we'll still be using the logic from the sirsi config.

### DIFF
--- a/lib/folio_record.rb
+++ b/lib/folio_record.rb
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/module/delegation'
+require 'active_support/core_ext/enumerable'
 require_relative 'traject/common/constants'
 require_relative 'locations_map'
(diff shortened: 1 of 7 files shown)


jcoyne (changes requested):
Before we merge this can we go over #861 and determine whether that approach will/won't work and why?  Otherwise I don't see a way of fixing #766.  If we do merge #861 then this PR may need updating.

thatbudakguy:
Per pairing session today I think we determined that the approach in #861 will work; moving this to draft until it can be rebased/fixed after merging #861

jcoyne (approved):

PR #880, opened 2023-05-16
Apply a schema to the holding model

jcoyne:
This helps us enforce the shape of the data and highlights the fields we think are important.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,3 +41,5 @@ group :deployment do
 end
 
 gem 'activesupport', '~> 7.0'
+
+gem 'dry-struct', '~> 1.6'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -80,6 +80,11 @@ GEM
       dry-logic (>= 1.4, < 2)
       dry-types (>= 1.7, < 2)
       zeitwerk (~> 2.6)
+    dry-struct (1.6.0)
+      dry-core (~> 1.0, < 2)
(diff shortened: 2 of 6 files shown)


cbeer:
I'm pretty -0 to this. It feels like a pretty complicated solution where either not using `fetch` or providing a default argument would seem to do just as well.

I also think there's value in keeping our code pretty close to the FOLIO data structure for talking to our colleagues in libsys + metadata about where/how we're mapping data. The abstraction seems like forces a deeper understading of the code, doesn't it?

jcoyne:
> I also think there's value in keeping our code pretty close to the FOLIO data structure for talking to our colleagues in libsys + metadata about where/how we're mapping data. 

Indeed. I did this by looking at the RAMLs that folio publishes. These data structures are derived from that definition.

```
quicktype -s schema ramls/holdingsrecord.json --namespace FolioTypes -o types.rb
```

cbeer:
Interesting. I feel like I need to know more about what it's doing.. the camel case -> snake case is fine, but `suppressFromDiscovery` turning into `discovery_suppress`, `location/effectiveLocation` getting collapsed, etc surprise me.

jcoyne:
`suppressFromDiscovery` was not in the raml model. That is something we extract from the jsonb in the query, so I had to add that myself.  I collapsed effectiveLocation as that is the only location that we're currently using.

PR #886, opened 2023-05-18
Provide more context in error reports

jcoyne:


lib/traject/config/sirsi_config.rb
```diff
@@ -62,8 +62,11 @@
 
   provide 'solr_writer.commit_on_close', true
   provide 'mapping_rescue', (lambda do |context, e|
-    Honeybadger.notify(e, context: { record: context.record_inspect, index_step: context.index_step.inspect })
-
+    Honeybadger.notify(e, context: {
+      context: context.record_inspect,
+      record: begin context.source_record.to_honeybadger_context rescue nil end,
```
cbeer, line 67: Some records can be pretty big (like, a small record is 50KB, and any moderately complex serial is going to be 1MB+). Is there a subset of information (that might be easier to read over in honeybadger) that might be helpful? We're already getting the catkey and can always look up the whole record in FOLIO anyway.

jcoyne: That's why I made this a method, which makes it easy to adjust what is returned.

lib/folio_record.rb
```diff
@@ -139,6 +139,10 @@ def as_json
     record
   end
 
+  def to_honeybadger_context
+    { hrid: }
```
cbeer, line 143:
```suggestion
    { hrid:, instance_id: }
```

We already have (essentially) the hrid from traject's `record_inspect` output, but no harm duplicating it I suppose. Maybe the UUID would be nice to include?

PR #889, opened 2023-05-18
Don't create mhld_display when there are no holdings statements

### LINKED ISSUE #884: [searchworks_traject_indexer/stage] NoMethodError: undefined method `fetch' for nil:NilClass
## Backtrace

    line 18 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: block in build
    line 17 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: map
    line 17 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: build

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/95841937)

jcoyne:
Fixes #884

### DIFF
--- a/lib/folio/mhld_builder.rb
+++ b/lib/folio/mhld_builder.rb
@@ -27,10 +27,13 @@ module Folio
 
     private
 
-    # Remove suppressed record and electronic records
+    # Remove suppressed records, electronic records, and records with no holdings statement
     def filtered_holdings
       holdings.filter_map do |holding|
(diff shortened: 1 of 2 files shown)


lib/folio/mhld_builder.rb
```diff
@@ -30,7 +30,10 @@ def build
     # Remove suppressed record and electronic records
```
marlo-longley, line 30: # Remove suppressed records, electronic records, and records with no holdings statement

PR #891, opened 2023-05-18
Manage loading with zeitwerk

jcoyne:
So we don't need to have require all over


lib/traject/config/sdr_config.rb
```diff
@@ -26,19 +17,19 @@ def log_skip(context)
 indexer = self
 
 settings do
-  provide 'writer_class_name', 'Traject::SolrBetterJsonWriter'
+  provide 'writer_class_name', 'Traject::Writers::SolrBetterJsonWriter'
```
cbeer, line 20: I think it'd be really nice to avoid injecting this 3rd level of class hierarchy. Did you try collapsing at all?

https://github.com/fxn/zeitwerk/blob/main/README.md#collapsing-directories

jcoyne: Naw, I thought it was better to use the conventions, but I can flatten.

cbeer: Thanks; I think traject's conventions and zeitwerks are in conflict. I think we'll be happier to keep with traject's for now.

jcoyne: I've reverted to use Traject's directories without modules.

thatbudakguy (approved):

PR #892, opened 2023-05-18
Handle YYYY formatted cronology

### LINKED ISSUE #885: [searchworks_traject_indexer/stage] Date::Error: invalid date
## Backtrace

    line 76 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: parse
    line 76 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: block in latest_received
    line 76 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/96071098)

jcoyne:
Fixes #885

### DIFF
--- a/lib/folio/mhld_builder.rb
+++ b/lib/folio/mhld_builder.rb
@@ -78,7 +78,7 @@ module Folio
     # @return [String] the latest received piece for a holding
     def latest_received(holding_id)
       # NOTE: We saw some piece records without 'chronology'. Was this just test data?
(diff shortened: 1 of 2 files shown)


cbeer:
Are there other chronology formats? Do we need to be even more defensive (or is sorting by chronology at all a fools errand)?

jcoyne:
I've asked, but we're waiting for a reply.  Currently this works for the records we've tried, so it seems an incremental improvement over what's currently in main.

PR #894, opened 2023-05-19
Create a SirsiHolding for FOLIO Bound-withs so that SearchWorks doesn't think they are On Order

marlo-longley:
- Create a SirsiHolding for FOLIO Bound-withs so that SearchWorks doesn't think they are On Order


lib/folio_record.rb
```diff
@@ -80,12 +80,35 @@ def sirsi_holdings
...
+  # TODO: remove this when we stop using SirsiHoldings
+  def bound_with_holdings
+    @bound_with_holdings ||= holdings.filter { |holding| holding.dig('holdingsType', 'name') == 'Bound-with' }.map do |holding|
+      parent_item = record['boundWithParents'].find { |parent| parent['childHoldingId'] == holding['id'] }
+      parent_item_location_code = parent_item.dig('parentItemLocation', 'permanentLocation', 'code')
+      library_code, home_location_code = LocationsMap.for(parent_item_location_code)
+      _current_library, current_location = LocationsMap.for(parent_item.dig('parentItemLocation', 'effectiveLocation', 'code'))
+      SirsiHolding.new(
+        call_number: holding['callNumber'],
+        current_location:,
```
thatbudakguy, line 103: I think we have the data for this — it should look the same as line 80:
```suggestion
        current_location: (current_location unless current_location == home_location_code),
```

lib/folio_record.rb
```diff
@@ -80,12 +80,35 @@ def sirsi_holdings
...
+        home_location: home_location_code,
+        library: library_code,
+        scheme: call_number_type_map(holding.dig('callNumberType', 'name')),
+        type: '',
+        barcode: '',
+        public_note: '',
+        tag: ''
```
thatbudakguy, line 110: what's the impact of not including these (especially barcode)? won't we need to render an item-level request link for boundwiths, which would need a barcode?

thatbudakguy (changes requested):

lib/folio_record.rb
marlo-longley: Good point, though I wasn't able to make much progress on the request button without circ rules. For this ticket trying to prevent "On Order" status felt like an actionable chunk. But I'm happy to try to populate these fields here if that makes most sense.

thatbudakguy: I feel like we may as well use these fields from the child, and see what happens? At least for testing, it might be good for the time being.

This is one of those cases where I think trying to get parity is a _bad_ idea; I think we saw that in current production, you get a location-level request link for the parent...which would make a librarian have to do the work to connect them. So, let's try something new here and see what happens.

marlo-longley: Sounds good. For me this raised questions about naming ("parent_barcode"?), SW design, etc., that were beyond what's needed to satisfy this particular ticket about preventing the On Order status.

thatbudakguy: That's fair — I think the underlying question of the ticket is "if not on-order, how should they appear?", and the important corollary is "they're not going to appear the same as how they do in current production". Preventing on-order status means that they still need to show as something else, so I think we should go for the more predictable thing here and not leave fields blank.

thatbudakguy: What would parity be? I think we saw that in existing symphony records, bound-with children have items, but in FOLIO they don't. It doesn't seem like there is a way to achieve parity in that situation.

thatbudakguy: @marlo-longley I guess we just need to go back and look at the `item_display` field for a bound-with in current production and double-check where each of the values comes from, and then try to recreate that. it might be possible...

marlo-longley:
I'm realizing this PR needs more investigation in 2 areas:
1. using the parent item's location in this SirsiHolding mashup removes the possibility of ever getting SEE-OTHER as a location, and therefore generating links such as the hated 'see parent' ones below. Instead we get "Available". This is not parity, and it IS the direction we want to move, so I think it's OK -- it will make more sense once we implement request buttons.
Screenshot 2023-05-19 at 16 28 47
Screenshot 2023-05-19 at 16 29 04


3. the child call number in production is `630.654 .I39M V.5:NO.2`. Using data from our DB query, the displayed call number resulting from this PR as is, is `630.654 .I39M`. Need to fix this.

marlo-longley:
This is now in a good place for review, with tests. I broke out the call number issue to #904

lib/folio_record.rb
```diff
@@ -80,12 +80,36 @@ def sirsi_holdings
...
+      SirsiHolding.new(
+        call_number: holding['callNumber'],
+        scheme: call_number_type_map(holding.dig('callNumberType', 'name')),
+        tag: {},
+        # parent item's barcode
+        barcode: parent_item['parentItemBarcode'],
+        # parent item's current location or SEE-OTHER (SAL3)
+        current_location: library_code == 'SAL3' ? '' : (current_location unless current_location == home_location_code),
+        # parent item's permanent location or SEE-OTHER (SAL3)
+        home_location: library_code == 'SAL3' ? 'SEE-OTHER' : home_location_code,
```
thatbudakguy, line 109: what's different about SAL3?

corylown: Cataloging practice: https://consul.stanford.edu/display/MD/Bound+withs

> For bound-with items, the Symphony child record's Home and Current Locations should be the same as the parent record's permanent Home Location. Child records should NEVER have location INPROCESS. EXCEPTION: When the bound-with item is in SAL3, both the Home and Current Locations on the child records should always be SEE-OTHER.

jcoyne: It would be good to put this comment in the code for future readers who may ask the same question.

marlo-longley:
OK, I added the comment about SAL3.

thatbudakguy (approved):

PR #916, opened 2023-06-06
Install foreman as part of the deploy.

cbeer:
Replaces #915 

Foreman says it shouldn't be added to the Gemfile (although I'm not personally convinced?), but it seems like we should want it installed as part of the deploy to ensure it is available.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -64,6 +64,12 @@ task :jruby_bundle_install do
   end
 end
 
+task :install_foreman do
+  on roles(:app) do
+    execute :gem, 'install', 'foreman', '-v', '~> 0.87.0'
+  end
+end
+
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

config/deploy.rb
```diff
@@ -74,3 +80,4 @@
 end
 
 before 'bundler:install', 'jruby_bundle_install'
+after 'bundler:install', 'install_foreman'
```
jcoyne, line 83: Could this be `before 'deploy:update_systemd_scripts', 'install_foreman'`?  We don't need it unless we're updating the systemd config.

cbeer: I think we always want it, in case someone wants to do it manually?

PR #920, opened 2023-06-09
Further shard the postgres query id space for better throughput

cbeer:
Data loading seems like it's been unhappy for awhile. It seems like the folio-test environment is likely highly resource constrained without enough scratch space to deliver even 1/16 of the holdings data. Bumping things up to 65535 shards seems like it gets things moving again (although a smaller value might also suffice... 0xff wasn't enough, but 0xffff seems to be 🤷‍♂️ ) 

The start-up cost for each query seems a little high, but maybe that's something we can tune in the future.


script/process_folio_postgres_to_kafka.rb
```diff
@@ -38,9 +38,21 @@
...
+    rescue PG::Error => e
+      raise(e) if attempts < 5
+
+      attempts += 1
+      Utils.logger.info e.message
+      sleep rand((2**attempts)..(2 * (2**attempts)))
+      retry
+    end
+
+    0
```
jcoyne, line 57: Doesn't this screw up the `counts`?

jcoyne (changes requested):

jcoyne (approved):

PR #929, opened 2023-06-14
Derive MARC 856 field data from FOLIO electronic access statements

### LINKED ISSUE #921: Handle Lane electronic holdings urls, stored in FOLIO holdings
Currently Lane maintains their URLs in MARC holdings records in Voyager. They have an automated process that copies the urls from MARC holdings records to MARC bib records every time a record is touched by Lane staff. 

In FOLIO, Lane would like to discontinue this automated record massaging process because:

- It will be confusing for staff to know where to maintain urls and data will diverge
- A similar process in FOLIO is conceptually possible but completely untested and…FOLIO factor
- Lane's desire to utilize the electronic access field in FOLIO holdings records is a more elegant solution in the new FOLIO framework.

In order to expose their electronic access urls from FOLIO holdings, SearchWorks will need to:

1. Look in FOLIO holdings records for an entry in Electronic access. If one exists use it. (Do this step regardless of Institution/Library)
2. IF there is no FOLIO holdings record with an Electronic access url, use the 856|u. 

In this order, we should be able to appropriately handle processes (like at Lane) which add the url to the FOLIO designated electronic access field, but can fall back to the method we've used historically in SUL (i.e. 856|u)

cbeer:
Fixes #921 

Lane is currently storing their electronic access in a MARC MHLD record. In FOLIO, they'd like to use FOLIO's native electronic access statements stored on their holdings record. We can ignore any data in the MARC bib records (so we don't have to have complicated merge logic).

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -23,6 +23,8 @@ settings do
     provide 'reader_class_name', 'Traject::FolioReader'
     provide 'folio.client', FolioClient.new(url: self['okapi.url'] || ENV.fetch('OKAPI_URL', nil))
   end
+
+  provide 'skip_empty_item_display', -1
 end
 
 ##
@@ -222,4 +224,12 @@ end
 
 to_field 'bound_with_parents_struct' do |record, accumulator|
   accumulator << record.bound_with_parents.to_json
+end
+
+##
+# Skip records for missing `item_display` field
+each_record do |record, context|
+  if context.output_hash['item_display'].blank? && context.output_hash['url_fulltext'].blank? && record.bound_with_parents.blank?
+    context.skip!('No item_display or url_fulltext field')
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -223,3 +225,11 @@ def holdings(record, context)
...
   accumulator << record.bound_with_parents.to_json
 end
+
+##
+# Skip records for missing `item_display` field
+each_record do |record, context|
+  if context.output_hash['item_display'].blank? && context.output_hash['url_fulltext'].blank? && record.bound_with_parents.blank?
+    context.skip!('No item_display or url_fulltext field')
+  end
+end
```
thatbudakguy, line 235: How's this work for on-order stuff? Doesn't that not have items?

thatbudakguy (approved):

lib/traject/config/folio_config.rb
cbeer: Some do, some don't. The ones that don't have items get a placeholder `item_display`:

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/sirsi_config.rb#L2398

PR #952, opened 2023-07-03
Separate data retrieval from updated date filtering

### LINKED ISSUE #951: FOLIO postgres query filters out too much data
The postgres query we're using to detect changed objects is also filtering out necessary data for indexing. E.g. when looking for changed items, we apply this condition:

```
sul_mod_inventory_storage.strtotimestamp((item.jsonb -> 'metadata'::text) ->> 'updatedDate'::text) > '#{@updated_after}'
```

But this means items with an updated date before the last run end up excluded from the returned items. We need to separate the filtering from the selected data to make sure we get all holdings, item, etc for an item.

cbeer:
Fixes #951

### DIFF
--- a/lib/traject/readers/folio_postgres_reader.rb
+++ b/lib/traject/readers/folio_postgres_reader.rb
@@ -32,8 +32,17 @@ module Traject
 
         # declare a cursor
         queries = if @updated_after
-                    conditions = %w[vi hr item cr cl cc].map { |table| "sul_mod_inventory_storage.strtotimestamp((#{table}.jsonb -> 'metadata'::text) ->> 'updatedDate'::text) > '#{@updated_after}'" }.map { |q| [q, @sql_filters].compact }
-                    conditions.map { |c| sql_query(c) }.join(') UNION (')
+                    filter_join = {
+                      'hr_filter' => 'LEFT JOIN sul_mod_inventory_storage.holdings_record hr_filter ON hr_filter.instanceid = vi.id',
+                      'item_filter' => 'LEFT JOIN sul_mod_inventory_storage.item item_filter ON item_filter.holdingsrecordid = hr.id',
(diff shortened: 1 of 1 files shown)


lib/traject/readers/folio_postgres_reader.rb
```diff
@@ -32,8 +32,15 @@ def each
...
-                    conditions.map { |c| sql_query(c) }.join(') UNION (')
+                    filter_join = {
+                      'hr_filter' => 'LEFT JOIN sul_mod_inventory_storage.holdings_record hr_filter ON hr_filter.instanceid = vi.id',
+                      'item_filter' => 'LEFT JOIN sul_mod_inventory_storage.item item_filter ON item_filter.holdingsrecordid = hr.id'
+                      'cr_filter'=> 'LEFT JOIN sul_mod_courses.coursereserves_reserves cr_filter ON (cr_filter.jsonb ->> \'itemId\')::uuid = item.id'
+                      'cl_filter'=> 'LEFT JOIN sul_mod_courses.coursereserves_courselistings cl_filter ON cl_filter.id = cr.courselistingid'
+                      'cc_filter'=> 'LEFT JOIN sul_mod_courses.coursereserves_courses cc_filter ON cc_filter.courselistingid = cl.id'
+                    }
+                    conditions = %w[vi hr_filter item_filter cr_filter cl_filter cc_filter].map { |table| "sul_mod_inventory_storage.strtotimestamp((#{table}.jsonb -> 'metadata'::text) ->> 'updatedDate'::text) > '#{@updated_after}'" }.map { |q| [q, @sql_filters].compact }
```
cbeer, line 42: Does `HAVING` work (performantly) for the child groupings instead of adding an additional join?

cbeer: I tried a couple variations of HAVING and all of them seemed to be much slower (15s vs minutes). Maybe duplicating the join conditions isn't so bad.

jcoyne (approved):

PR #957, opened 2023-07-05
Map SUL/SDR location to SUL/INTERNET for parity

### LINKED ISSUE #938: Map SUL/SDR location to SUL/INTERNET for parity
Some electronic holdings in FOLIO have the location code SUL-SDR, but in Symphony were in SUL/INTERNET.

jcoyne:
Fixes #938

### DIFF
--- a/lib/locations_map.rb
+++ b/lib/locations_map.rb
@@ -2,6 +2,7 @@
 
 require 'csv'
 
+# This starts with the location.tsv file provided by libsys. We make any modifications to that data here.
 class LocationsMap
   include Singleton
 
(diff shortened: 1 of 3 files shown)


lib/folio/eresource_holdings_builder.rb
```diff
@@ -85,7 +85,8 @@ def barcode(index)
     end
 
     def home_location_code
```
cbeer, line 87: Should we push this into `LocationsMap`, where we're already doing some of this location code massaging?

jcoyne: How about into locations.tsv?

cbeer: I thought we were trying to keep our locations.tsv in sync with what libsys is using.

lib/translation_maps/locations.tsv
```diff
@@ -449,4 +449,4 @@ SSRC-FIC-S	GREEN	GRE-DOCS-MICROFICHE
 SDR	LAW	LAW-SDR
 SDR	BUSINESS	BUS-SDR
 SDR	HOOVER	HILA-SDR
-SDR	SUL	SUL-SDR
+INTERNET	SUL	SUL-SDR
```
cbeer, line 452: (Copied from the outdated thread)

I thought we were trying to keep our locations.tsv in sync with what libsys is using.

jcoyne: 🤷 are we documenting that somewhere?

cbeer: Sure, anywhere you think would be appropriate. It seems like a very good idea not to change that upstream data.

cbeer (approved):

PR #967, opened 2023-07-11
Fix bad autocorrecting

cbeer:


lib/traject/writers/solr_better_json_writer.rb
```diff
@@ -79,7 +79,7 @@ def send_single(c)
       resp = @http_client.post @solr_update_url, json_package, 'Content-type' => 'application/json'
       # Catch Timeouts and network errors as skipped records, but otherwise
       # allow unexpected errors to propagate up.
-    rescue *skippable_exceptions => e
+    rescue *skippable_exceptions => exception # rubocop:disable Naming/RescuedExceptionsVariableName https://github.com/rubocop/rubocop/issues/11809
```
jcoyne, line 82: Wasn't this fixed awhile back? Do we still need to disable it?

cbeer: 🤷‍♂️ I tried removing them and rubocop didn't like it

jcoyne: Ah, I see.

jcoyne (approved):

lib/traject/writers/solr_better_json_writer.rb
jcoyne: I created https://github.com/rubocop/rubocop/issues/12042 to argue the point.

PR #971, opened 2023-07-12
Refine FOLIO MHLD enumeration + chronology sort for calculating lates…

### LINKED ISSUE #905: Refine FOLIO MHLD enumeration + chronology sort for calculating latest received
Symphony gave us internally-generated serial control data that let us properly sort data. FOLIO just has free-text entries for `enumeration` and `chronology` (both or either of which are optional) e.g.:

| enumeration | chronology |
|--------------|--------------|
| v.19 |  |
| v.14 | 2019 |
|(v.) 8 | 2014 |
|v. 48 fasc. 1 | |
|v. 48 fasc. 2 | | 
| v.301 pt.4/6 | Feb 2021 | 
|jahrg.63 nr.9/10 | Sep/Oct 2022 |
|v.58 | Jan/Dec 2022|
| t.77 no.884/885 | Jan/Feb 2021 |
| dec/jan 2018/2019 | |
| | 2013 |
|v. 53 no. 1/4 | Spr/Win 2012 |
|v.40/41 | 2016/2017 |
|v.47 no.1 =no.185 | Jan/Mar 2019|
| jahrg.33 nr.31/32 | Jul/Aug 27/3, 2018 |
| jahrg.36 nr.7 | Feb 12, 2021 |
|mar 31 | 2016|
| gift - do not predict | do not claim |

(See the full list at https://docs.google.com/spreadsheets/d/1W7cceLRVy4Qg1LXeoT7JT62STuu6Ouh9/edit#gid=18684243)

Extracting and parsing the various date formats in chronology seems like a hopeless exercise, but perhaps we can get to "good enough" with some common cases and falling back to extracting numeric parts and hoping for the best

Proposal:

1) If there's either a chronology or a chronology + enumeration, sort first by a chronology data, either by: 

a) Extracting a valid month + year (e.g. `Jun 2018`) from the chronology. There may be other Information in the field as well (e.g. `Dec/Jan 2022/20233`). If there's other garbage that make parsing difficult, ignore the month and fall back on (b). 
 
b) Extracting the last valid year from chronology (if one is available). A year should have 4 digits starting 19xx or 20xx. The chronology may have other text in it, but maybe we can ignore it. There may be other numeric data in the field (e.g. `2013/2014`, `May 31, 2013`, `2017 No. 4/5/5`). 

c) Sort second by the numeric tokens in the enumeration. There's a lot of variation in the text content, but many times it seems like there's some numbers. E.g.

```
v.86 => [86]
v. 63 no. 1 => [63, 1]
2020/2021 => [2020, 2021]
bd.54 t.2 => [54, 2]
t.78 no.907 => [78, 907]
```

d) Sort last by the text value of the enumeration

2) If there's only an enumeration:

a) check if there's a reasonable looking year(20xx or 19xx) or month year and treat it as a chronology (as in 1a + 1b)
b) apply 1c + 1d

jcoyne:
…t received

Fixes #905

### DIFF
--- a/lib/folio/holdings.rb
+++ b/lib/folio/holdings.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module Folio
+  class Holdings
+    YEAR_REGEX = /(?:19|20)\d{2}/
+    # We've seen chronologies that looks like full dates, MON YYYY, WIN YYYY, and nil
+    def self.find_latest(holdings)
+      # NOTE: We saw some piece records without 'chronology'. Was this just test data?
+      pieces = holdings.filter_map do |piece|
+        piece.merge(sortable_date: sortable_date(piece),
+                    sortable_tokens: sortable_tokens(piece))
+      end
+      pieces.max_by { |piece| [piece.fetch(:sortable_date), piece.fetch(:sortable_tokens), piece['enumeration']] }
+    end
+
+    # @return [Date] a date derived from chronology, enumeration or a really old date (sorts to back) if none are found
+    def self.sortable_date(piece)
+      extract_sortable_date(piece['chronology']) ||
+        extract_sortable_date(piece['enumeration']) ||
+        Date.parse('1000-01-01')
+    end
+
+    # @return [Array<Integer>] a list of integer tokens derived from chronology and enumeration.
+    def self.sortable_tokens(piece)
+      Array(
+        extract_sortable_tokens(piece['chronology']) ||
+          extract_sortable_tokens(piece['enumeration'])
+      )
+    end
+
+    # Find any dates that are present in the fields
+    def self.extract_sortable_date(date_str)
+      return unless date_str
+
+      begin
+        return Date.parse(date_str) # handles MON YYYY
+      rescue Date::Error
+        nil
+      end
+
+      return unless (years = date_str.scan(YEAR_REGEX).presence)
+
+      max_year = years.map(&:to_i).max
+      Date.parse("#{max_year}-01-01")
+    end
+
+    # Look for other tokens, which are not dates, that can be used to sort.
+    # We need to exclude dates so this field can be used as a tie-breaker
+    def self.extract_sortable_tokens(str)
+      return unless str.present?
+
+      tokens = str.scan(/\d+/).grep_v(YEAR_REGEX)
+      tokens.map(&:to_i) if tokens.present?
+    end
+  end
+end
--- a/lib/folio/mhld_builder.rb
+++ b/lib/folio/mhld_builder.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../locations_map'
+require_relative 'holdings'
 
 module Folio
   class MhldBuilder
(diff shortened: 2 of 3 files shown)


lib/folio/holdings.rb
```diff
@@ -0,0 +1,41 @@
...
+      Date.parse("#{max_year}-01-01")
+    end
+
+    # Look for other tokens, which are not dates, that can be used to sort.
+    # We need to exclude dates so this field can be used as a tie-breaker
+    def self.extract_sortable_tokens(str)
+      return unless str
+
+      tokens = str.scan(/\d+/).grep_v(YEAR_REGEX)
+      tokens.map(&:to_i).max
```
cbeer, line 38: In #905, one of the examples was "bd.54 t.2" (say, vs "bd.54 t.1'). Won't the `max` here just pull out `54` and indeterminately sort the rest?

lib/folio/holdings.rb
```diff
@@ -0,0 +1,41 @@
...
+  class Holdings
+    YEAR_REGEX = /(?:19|20)\d{2}/
+    # We've seen chronologies that looks like full dates, MON YYYY, WIN YYYY, and nil
+    def self.find_latest(holdings)
+      # NOTE: We saw some piece records without 'chronology'. Was this just test data?
+      pieces = holdings.filter_map do |piece|
+        piece.merge(sortable_date: extract_sortable_date(piece['chronology']) || extract_sortable_date(piece['enumeration']),
+                    sortable_tokens: extract_sortable_tokens(piece['chronology']) || extract_sortable_tokens(piece['enumeration']))
+      end
+      pieces.max_by { |piece| [piece.fetch(:sortable_date), piece.fetch(:sortable_tokens)] }
```
cbeer, line 13: In #905, I suggested step d to fall back on a text sort. Was that problematic, and if so, could you drop a comment on #905 with why?

lib/folio/holdings.rb
jcoyne: Ah, good catch. I've added a test for this case.

lib/folio/holdings.rb
jcoyne: I've added a fallback to text sort now.

lib/folio/holdings.rb
```diff
@@ -0,0 +1,41 @@
...
+      # NOTE: We saw some piece records without 'chronology'. Was this just test data?
+      pieces = holdings.filter_map do |piece|
+        piece.merge(sortable_date: extract_sortable_date(piece['chronology']) || extract_sortable_date(piece['enumeration']),
+                    sortable_tokens: Array(extract_sortable_tokens(piece['chronology']) || extract_sortable_tokens(piece['enumeration']) || piece['enumeration']))
+      end
+      pieces.max_by { |piece| [piece.fetch(:sortable_date)] + piece.fetch(:sortable_tokens) }
+    end
+
+    # Find any dates that are present in the fields
+    def self.extract_sortable_date(date_str)
```
cbeer, line 17: I think it's possible we'd have some pieces with a sortable date and some without:

```
> [[Date.parse('2023-01-1'), nil], [nil, nil]].sort
(irb):5:in `sort': comparison of Array with Array failed (ArgumentError)
```

Should this method always return a date object?

lib/folio/holdings.rb
```diff
@@ -0,0 +1,57 @@
...
+        extract_sortable_date(piece['enumeration']) ||
+        Date.parse('1000-01-01')
+    end
+
+    # @return [Array<String>] a list of tokens derived from chronology and enumeration. Numbers are favored over strings.
+    def self.sortable_tokens(piece)
+      Array(
+        extract_sortable_tokens(piece['chronology']) ||
+          extract_sortable_tokens(piece['enumeration']) ||
+          piece['enumeration']
```
cbeer, line 28: Isn't there still a problem if we try to sort between the "sortable tokens" and strings?

```
[Array('1'), Array(1)].sort
(irb):4:in `sort': comparison of Array with Array failed (ArgumentError)
```

jcoyne: Okay, I've added that test and made it pass.

PR #972, opened 2023-07-12
Only retrieve ACTUAL source records.

cbeer:
FOLIO keeps (all? some?) versions of the source records around.


marlo-longley (approved):
Lol, I thought ACTUAL was just your emphasis for the title. But it's real.

PR #986, opened 2023-07-18
Find items if the locations have been updated

jcoyne:


jcoyne:
NOTE, this pushed the query over the 3min mark ( in `spec/lib/traject/readers/folio_postgres_reader_spec.rb`), so we may need a different approach.

PR #987, opened 2023-07-18
Write a test that shows that the query is valid

jcoyne:
This also gives us a way to benchmark in order to see if we've adversely impacted the performance of the query.


cbeer:
👍 I like it. Should we fail the build if the test takes over... 3 minutes?

PR #990, opened 2023-07-18
Strip subfield 0 data that starts with (SIRSI)

cbeer:
These aren't present in the Symphony MARC record and make looking for indexing differences annoying.


jcoyne (commented):
Can you please add a test?

cbeer:
Sure, although `compare_sirsi_and_folio` is a pretty good test for this.

cbeer:
Hopefully this logic is only as permanent as our interesting in comparing data against Symphony 🤷‍♂️

PR #991, opened 2023-07-18
Use filter_map when generating holdings

cbeer:
Regression from #985 now that the `map` can break early.


jcoyne (commented):
should there be a test?

PR #993, opened 2023-07-20
Use library labels from Folio as source for the building_facet values

### LINKED ISSUE #896: Use FOLIO's discovery labels
There's also an opportunity to use labels from various tables instead of maintaining a code->user label mapping in Searchworks:
- [ ] `sul_mod_inventory_storage.location`  
- [ ] `sul_mod_inventory_storage.loclibrary`

If we can use those mappings, I think we have a choice whether to fetch them at query time or store them in this project and update them regularly.

If we fetch them at query time, it'll be important for consistency to reindex the instance data if/when the labels change (e.g. to avoid having the same library show up multiple times in the library facet (`building_facet` field))

If we decide to stash them in the project, we'll need a process for retrieving the data and kicking off a full reindex.

jcoyne:
Fixes #896


README.md
```diff
@@ -188,3 +188,13 @@ export DATABASE_URL=postgres://[user]:[password]@localhost/okapi
 ```sh
 ./script/download_folio_record.rb a123456 > record.json
 ```
+
+#### Synching data
+##### Libraries
+If the name of a library in Folio has changed, you'll want to export the list of libraries with their labels and check it in here.  You can do this using the psql copy command.
+
+```
+\copy (select id, jsonb ->>'code' as code, jsonb->>'name' as name from loclibrary) to 'config/folio/loclibrary.csv' csv header;
```
cbeer, line 197: I guess I imagined we'd have a script that stores the full JSON responses like we're doing in Searchworks + sul-requests, or we'd generate something `Traject::TranslationMap` could use natively. Going directly to the database seems unnecessary and would make it harder to wire up to Jenkins (at the very least, we'd need to get the postgres client libraries installed over there 🤷‍♂️ )

jcoyne: I've added https://github.com/sul-dlss/searchworks_traject_indexer/pull/1044 to get that data.

lib/libraries_map.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Map library codes to labels exported from Folio
+class LibrariesMap
+  def self.for(key)
+    return if key == 'SUL' # do not index building_facet on electronic resources
+
+    Folio::Types.libraries.values.find { |library| library.fetch('code') == key }.fetch('name')
```
cbeer, line 8: This doesn't work for Lane's material:

```
./lib/libraries_map.rb:8:in `for': undefined method `fetch' for nil:NilClass (NoMethodError)

    Folio::Types.libraries.values.find { |library| library.fetch('code') == key }.fetch('name')
                                                                                 ^^^^^^
```

It looks like we're passing in the Symphony library codes (which makes sense... for parity), but the Symphony + FOLIO codes for Lane are different (`LANE` vs `LANE-MED`).

We should account for that difference here, and also add a reasonable fallback if we don't get a name (probably nil so it is ignored?). This is just populating a facet, so it's not the end of the world if we don't get a match for whatever reason.

lib/libraries_map.rb
```diff
@@ -0,0 +1,20 @@
...
+class LibrariesMap
+  def self.for(key)
+    return if key == 'SUL' # do not index building_facet on electronic resources
+
+    key = 'LANE' if key == 'LANE-MED' # If symphony codes are passed as input, we need to update to the folio code
+
+    library = Folio::Types.libraries.values.find { |candidate| candidate.fetch('code') == key }
+    return library.fetch('name') if library
+
+    Honeybadger.notify('Lookup for library not found', context: { key: })
```
cbeer, line 13:
```suggestion
```
This seems like to swamp honeybadger if we're wrong, and fixing it is outside of our control (and ought to happen when we receive updated library information from FOLIO, if things haven't gone incredibly wrong)

jcoyne: @cbeer  Can you explain "ought to happen"? I'm having trouble picturing the situation where we don't have this data.  That seems like something worth documenting.

cbeer: If someone adds a new library code and starts using it, but it isn't in that cached list of libraries (that we only updated daily)...?

PR #1005, opened 2023-07-25
Inject locations after the query for way better performance

cbeer:
#998 introduced a query performance regression with the additional location lookup. 

Pulling the location data out of the query entirely seems to halve the query time for our test case (18s -> 9s).


jcoyne (approved):
Great idea!

PR #1007, opened 2023-07-25
Don't return a holding when there is a migration error

### LINKED ISSUE #1001: [searchworks_traject_indexer/stage] NoMethodError: undefined method `library' for nil:NilClass
## Backtrace

    line 490 of [PROJECT_ROOT]/lib/traject/config/sirsi_config.rb: block (3 levels) in load_config_file
    line 490 of [PROJECT_ROOT]/lib/traject/config/sirsi_config.rb: any?
    line 490 of [PROJECT_ROOT]/lib/traject/config/sirsi_config.rb: block (2 levels) in load_config_file

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/98558911)

jcoyne:
Previously if the location ended with MIGRATE-ERR, it would add a nil into the list of sirsi_holdings.  This later cause a NoMethodError when iterating over this list.  Since the nil is not meaningful, we can compact them from the array.

Fixes #1001

### DIFF
--- a/lib/folio_record.rb
+++ b/lib/folio_record.rb
@@ -99,7 +99,7 @@ class FolioRecord
         public_note: item['notes']&.map { |n| ".#{n['itemNoteTypeName']&.upcase}. #{n['note']}" }&.join("\n"),
         tag: item
       )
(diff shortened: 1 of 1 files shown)


jcoyne:
Duplicate of #991

PR #1009, opened 2023-07-25
Guard against nil items and holdings

### LINKED ISSUE #1008: [searchworks_traject_indexer/stage] NoMethodError: undefined method `[]' for nil:NilClass
## Backtrace

    line 126 of [PROJECT_ROOT]/lib/traject/readers/folio_postgres_reader.rb: block (4 levels) in each
    line 124 of [PROJECT_ROOT]/lib/traject/readers/folio_postgres_reader.rb: each
    line 124 of [PROJECT_ROOT]/lib/traject/readers/folio_postgres_reader.rb: block (3 levels) in each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/98572259)

cbeer:
Fixes #1008

It's not clear to me how this happens 🤷‍♂️

### DIFF
--- a/lib/traject/readers/folio_postgres_reader.rb
+++ b/lib/traject/readers/folio_postgres_reader.rb
@@ -225,7 +225,7 @@ module Traject
                             )
                           ELSE NULL END::jsonb
                   )
-                )
+                ) FILTER (WHERE hr.id IS NOT NULL), '[]'::jsonb
               ),
             'pieces',
               COALESCE(


jcoyne:
@cbeer can you add a sample HRID here so we can investigate?

cbeer:
Nope; the honeybadger report doesn't give us anything. All I know is the uuid probably starts with a 0 or 1 😬

cbeer:
I found one of the records; it's a instance without any holdings or items. Items were already filtered + coalesced, but we must have dropped the filter from upstream:
https://github.com/folio-org/mod-inventory-storage/blob/9b7b8fc31befc335629f8d38e13e647822f7c8c6/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql#L274

PR #1016, opened 2023-07-26
Handle holding records that have neither a chronology nor enumeration

### LINKED ISSUE #1013: [searchworks_traject_indexer/stage] KeyError: key not found: "chronology"
## Backtrace

    line 87 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: fetch
    line 87 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: latest_received
    line 27 of [PROJECT_ROOT]/lib/folio/mhld_builder.rb: block in build

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/98580232)

jcoyne:
Fixes #1013

### DIFF
--- a/lib/folio/mhld_builder.rb
+++ b/lib/folio/mhld_builder.rb
@@ -84,7 +84,7 @@ module Folio
       return unless latest_piece && order_is_ongoing_and_open?(latest_piece)
 
       enumeration = latest_piece['enumeration'] # may not be present
-      chronology = latest_piece.fetch('chronology')
+      chronology = latest_piece['chronology'] # may not be present
       enumeration ? "#{enumeration} (#{chronology})" : chronology
     end
 
--- a/spec/lib/folio/mhld_builder_spec.rb
+++ b/spec/lib/folio/mhld_builder_spec.rb
@@ -329,4 +329,87 @@ RSpec.describe Folio::MhldBuilder do
 
     it { is_expected.to eq ['GREEN -|- CURRENTPER -|- Latest issues in CURRENT PERIODICALS; earlier issues in STACKS -|- 2018 -|- JUN 2022'] }
   end
+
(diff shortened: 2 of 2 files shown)


lib/folio/mhld_builder.rb
```diff
@@ -84,7 +84,7 @@ def latest_received(holding_id)
       return unless latest_piece && order_is_ongoing_and_open?(latest_piece)
 
       enumeration = latest_piece['enumeration'] # may not be present
-      chronology = latest_piece.fetch('chronology')
+      chronology = latest_piece['chronology'] # may not be present
       enumeration ? "#{enumeration} (#{chronology})" : chronology
```
cbeer, line 88: Do we need to tweak the string output in that case?

```suggestion
      enumeration && chronology ? "#{enumeration} (#{chronology})" : enumeration || chronology
```

jcoyne: It currently looks to have parity with what is coming out of Symphony, so I see no reason to make adjustments there.

cbeer: I'm surprised we'd leave a blank `()` sitting around if there was an enumeration but not a chronology, but it's difficult to  deciphering the existing logic and/or translate the MHLD fields to FOLIO data.

Is this the equivalent code?
https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/mhld_field.rb#L57-L62

jcoyne: I don't know. This was just to fix the missing method error. If you think there's an error in the output, I would prefer that we make that a new issue, find an example record and add a test to show the behavior we want.

cbeer: 🤷‍♂️

PR #940, opened 2023-06-22
Set current location based on FOLIO item status; fixes #937

cbeer:

PR #1020, opened 2023-07-26
Set current location from request service point for 'awaiting pickup' items

### LINKED ISSUE #974: Handling awaiting + in-transit statuses with more specific information
In Symphony, we have `*-LOAN` locations (meaning e.g. at the Green Loan desk). We generally map those to "On hold for a borrower", but there are some library-specific exceptions (ARS, RUMSEY).

In #973, we mapped to a generic code, but we should see if there's anything more we can do to reach parity.

Work with Darsi as needed to make an item go through FOLIO paging to become "Awaiting Pickup"

corylown:
Closes #974


jcoyne (commented):
Can you confirm you've run this with the folio_postgres_reader_spec.rb and the time isn't adversely affected?

lib/folio/status_current_location.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    private
+
+    def status
+      item['status']
+    end
+
+    # ARS-LOAN, RUM-LOAN, and SPE-LOAN each receives a
+    # special label in SearchWorks. Other codes are generically
```
jcoyne, line 39: Where will these labels come from in Searchworks? Aren't we moving toward using Folio location labels?

jcoyne (commented):
Aren't we going to be moving away from item_display and towards using item_display_struct?   Do we need this change there as well?

corylown:
> Aren't we going to be moving away from item_display and towards using item_display_struct? Do we need this change there as well?

No additional change is needed. This is modifying the `current_location` value used to construct `sirsi_holdings` for FOLIO records and `sirsi_holdings` is what's used to construct the new`item_display_struct` field (which is then used to create the legacy item_display field).

lib/folio/status_current_location.rb
corylown: The equivalent in FOLIO for these locations would be service points, which I don't think we're doing anything with in SearchWorks yet. We might want to. But this gets us to parity with Symphony by relying on the existing `Constants:LOCS` mappings.

corylown:
> Can you confirm you've run this with the folio_postgres_reader_spec.rb and the time isn't adversely affected?

It's still within the expected range of 30 to 90 seconds:

```
Traject::FolioPostgresReader
  creates FolioRecords

Top 1 slowest examples (51.85 seconds, 100.0% of total time):
  Traject::FolioPostgresReader creates FolioRecords
    51.85 seconds ./spec/lib/traject/readers/folio_postgres_reader_spec.rb:19

Finished in 51.86 seconds (files took 1.57 seconds to load)
1 example, 0 failures
```

lib/traject/readers/folio_postgres_reader.rb
```diff
@@ -312,6 +345,9 @@ def sql_query(conditions, addl_from: nil)
           ON parentItem.holdingsRecordId = parentHolding.id
       LEFT JOIN sul_mod_inventory_storage.instance parentInstance
           ON parentHolding.instanceid = parentInstance.id
+      -- Requests relation
+      LEFT JOIN sul_mod_circulation_storage.request request
+          ON (request.jsonb ->> 'instanceId')::uuid = vi.id
```
cbeer, line 350: Does this retrieving every request for an instance? Can we further filter it by request status just to the open/waiting request?

corylown: Yeah, I think we should probably filter the requests to just the one's we're interested in.

lib/folio/status_current_location.rb
corylown: @cbeer I'm confused about what the plan is for `Constants::LOCS` vs FOLIO display names. My guess is we're still going to have location codes that don't map to anything in FOLIO and we'll need to maintain a fallback list in SearchWorks (maybe I'm wrong about this). This PR is a refinement to previous work to map folio statuses to symphony locations so they'll display right in searchworks. Is this no longer what we want to do?

See:

- https://github.com/sul-dlss/searchworks_traject_indexer/pull/940/files
- https://github.com/sul-dlss/searchworks_traject_indexer/pull/973/files

lib/traject/readers/folio_postgres_reader.rb
```diff
@@ -254,6 +276,17 @@ def sql_query(conditions, addl_from: nil)
                     'instructorObjects', cl.jsonb #> '{instructorObjects}'
                   )
                 ) FILTER (WHERE cc.id IS NOT NULL),
+              '[]'::jsonb),
+            'requests',
```
cbeer, line 280: 🤔 If we always have an item, could we nest this under the item (like we're doing with bound-withs and the holdings record)?

corylown: I think we could. I'm not sure how to write that query though.

lib/folio/status_current_location.rb
corylown: Tried to capture post-folio work related to this in #1026

PR #1037, opened 2023-07-31
Update dependencies

sul-devops-team:


lib/traject/readers/marc_combining_reader.rb
```diff
@@ -6,11 +6,11 @@ def initialize(left, right)
     @right = right
   end
 
-  def each(*args, &)
+  def each(...)
```
jcoyne, line 9: Isn't this a ruby 3.2 feature? I don't think the version of jruby we're using supports the 3.2 language spec.

PR #1042, opened 2023-08-02
Use item_display_struct to compare FOLIO + Symphony data

cbeer:


jcoyne:
It looks like this needs a rebase.

jcoyne (approved):

PR #1057, opened 2023-08-07
Handle nil call_numbers

### LINKED ISSUE #1049: [searchworks_traject_indexer/stage] NoMethodError: undefined method `match?' for nil:NilClass
## Backtrace

    line 189 of [PROJECT_ROOT]/lib/sirsi_holding.rb: valid_lc?
    line 78 of [PROJECT_ROOT]/lib/sirsi_holding.rb: bad_lc_lane_call_number?
    line 2182 of [PROJECT_ROOT]/lib/traject/config/sirsi_config.rb: block (3 levels) in load_config_file

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/57331/faults/98852854)

jcoyne:
This may happen for on-order items
Fixes #1049

### DIFF
--- a/lib/sirsi_holding.rb
+++ b/lib/sirsi_holding.rb
@@ -149,7 +149,7 @@ class SirsiHolding
   private
 
   # Call number normalization ported from solrmarc code
-  def normalize_call_number(call_number = '')
+  def normalize_call_number(call_number)
     return call_number unless %w[LC DEWEY].include?(call_number_type) # Normalization only applied to LC/Dewey
 
     call_number = call_number.strip.gsub(/\s\s+/, ' ') # reduce multiple whitespace chars to a single space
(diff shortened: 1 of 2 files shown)


lib/sirsi_holding.rb
```diff
@@ -149,7 +149,7 @@ def to_item_display_hash
   private
 
   # Call number normalization ported from solrmarc code
-  def normalize_call_number(call_number = '')
+  def normalize_call_number(call_number)
```
cbeer, line 152: Couldn't this `call_number` be nil too? Wouldn't lines 155-158 need some handling?

cbeer: Isn't it possible to have a call_number_type set, but end up with nil call number? I don't see `call_number_type` checking anything besides the provided schema.

jcoyne: Maybe, but we've never seen data like that. I doubt we ever will. When the call_number has been nil (situation we're fixing), the type is also nil.

PR #1059, opened 2023-08-07
Stop providing a default when the argument is required

jcoyne:


cbeer:
I added this in anticipation we'd want it for https://github.com/sul-dlss/searchworks_traject_indexer/issues/1028. Is that no longer the case?

jcoyne:
@cbeer i'm sorry I didn't understand that this will be in use. I've amended the PR to revert that.

PR #1060, opened 2023-08-07
folio crez extra info

thatbudakguy:
- Explicitly require FolioPostgresReader in download script
- Add reserve desk, loan period, and course ID to item_display for FOLIO


spec/fixtures/files/whodunit.json
thatbudakguy, line 1: I re-exported this from the database and tweaked the data slightly so that it would be a good basis for testing.

spec/lib/traject/config/folio_course_reserves_spec.rb
```diff
@@ -34,4 +34,21 @@
       'EDUC-147-01 -|- Stanford and Its Worlds: 1885-present -|- Mitchell Stevens'
     ]
   end
+
+  context 'for item display' do
```
jcoyne, line 38: I don't think this is a context (see https://www.betterspecs.org/#contexts).  How about this:
```suggestion
  describe 'the items in the item_display_struct field' do
```

spec/lib/traject/config/folio_course_reserves_spec.rb
```diff
@@ -34,4 +34,21 @@
...
+    # Filter to the item that's on reserve so we can check its item_display
+    subject(:item_display) { result['item_display_struct'].find { |item| item['barcode'] == '36105232609540' } }
+
+    it 'adds the course ID' do
+      expect(item_display['course_id']).to eq 'HISTORY-58E-01'
```
jcoyne, line 43: Could this all be one test with
```ruby
expect(item_display).to include('course_id' => 'HISTORY-58E-01',
                                'reserve_desk' => 'GRE-CRES',
                                'loan_period' => '2-hour loan')
```

jcoyne (changes requested):

lib/traject/config/folio_config.rb
```diff
@@ -197,6 +197,29 @@ def holdings(record, context)
   end
 end
 
+# We update item_display once we have crez info
+to_field 'item_display_struct' do |record, _accumulator, context|
+  next if record.courses.empty?
+
+  context.output_hash['item_display_struct'].map! do |item_display_str|
```
cbeer, line 204: Do we need to wait?  https://github.com/sul-dlss/searchworks_traject_indexer/pull/1029 opened up a seam for you to set this up when we're creating the holding.

spec/lib/traject/config/folio_course_reserves_spec.rb
thatbudakguy: I was looking at https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording, but didn't realize that the `for` shown in the example isn't actually a default value.

lib/traject/config/folio_config.rb
thatbudakguy: Wish I'd known that when I picked up this ticket! I'll try it that way.

spec/lib/traject/config/folio_course_reserves_spec.rb
thatbudakguy: Sure, but I like having the `it` declarations be more atomic. More granular tests is better to me.

lib/traject/config/folio_config.rb
thatbudakguy: Updated

lib/folio_record.rb
```diff
@@ -87,6 +87,19 @@ def sirsi_holdings
       _current_library, current_location = LocationsMap.for(item.dig('location', 'temporaryLocation', 'code'))
       current_location ||= Folio::StatusCurrentLocation.new(item).current_location
 
+      # TODO: we don't handle multiple courses for a single item, because it's beyond parity
+      item_courses = courses.select { |course| course[:listing_id] == item['courseListingId'] }
+      course = item_courses.first
```
cbeer, line 92:
```suggestion
      course = courses.first { |course| course[:listing_id] == item['courseListingId'] }
```

lib/folio_record.rb
```diff
@@ -87,6 +87,19 @@ def sirsi_holdings
       _current_library, current_location = LocationsMap.for(item.dig('location', 'temporaryLocation', 'code'))
       current_location ||= Folio::StatusCurrentLocation.new(item).current_location
 
+      # TODO: we don't handle multiple courses for a single item, because it's beyond parity
```
jcoyne, line 90:
```suggestion
      # NOTE: we don't handle multiple courses for a single item, because it's beyond parity with how things worked for Symphony
```

lib/folio_record.rb
```diff
@@ -87,6 +87,19 @@ def sirsi_holdings
...
+      item_courses = courses.select { |course| course[:listing_id] == item['courseListingId'] }
+      course = item_courses.first
+
+      # We use loan types as loan periods for course reserves so that we don't need to check circ rules
+      # Items on reserve in FOLIO usually have a temporary loan type that indicates the loan period
+      # "3-day reserve" -> "3-day loan"
+      course_reserves = {
```
jcoyne, line 97: Can we remove the conditional from line 114 and define `course_reserves = {}` just prior to this line?

lib/folio_record.rb
thatbudakguy: Did not know you could pass that a block! Cool!

spec/lib/traject/config/folio_course_reserves_spec.rb
thatbudakguy: OK, I can do that.

lib/folio_record.rb
thatbudakguy: Done

spec/lib/traject/config/folio_course_reserves_spec.rb
thatbudakguy: Done

jcoyne (approved):

PR #1061, opened 2023-08-08
Guard against nil call numbers

cbeer:
Continuation of #1057 ; we're guarding against nil call numbers in so many other places, it seems like we should do so here too.

### DIFF
--- a/lib/sirsi_holding.rb
+++ b/lib/sirsi_holding.rb
@@ -150,7 +150,7 @@ class SirsiHolding
 
   # Call number normalization ported from solrmarc code
   def normalize_call_number(call_number)
-    return call_number unless %w[LC DEWEY].include?(call_number_type) # Normalization only applied to LC/Dewey
+    return call_number unless call_number && %w[LC DEWEY].include?(call_number_type) # Normalization only applied to LC/Dewey
 
     call_number = call_number.strip.gsub(/\s\s+/, ' ') # reduce multiple whitespace chars to a single space
     call_number = call_number.gsub('. .', ' .') # reduce double periods to a single period
@@ -211,6 +211,8 @@ class SirsiHolding
     end
 
     def normalized_lc
+      return unless call_number
+
       call_number.gsub(/\s\s+/, ' ') # change all multiple whitespace chars to a single space
                  .gsub(/\s?\.\s?/, '.') # remove a space before or after a period
                  .gsub(/^([A-Z][A-Z]?[A-Z]?) ([0-9])/, '\1\2') # remove space between class letters and digits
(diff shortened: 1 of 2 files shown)


lib/sirsi_holding.rb
```diff
@@ -150,7 +150,7 @@ def to_item_display_hash
 
   # Call number normalization ported from solrmarc code
   def normalize_call_number(call_number)
-    return call_number unless %w[LC DEWEY].include?(call_number_type) # Normalization only applied to LC/Dewey
+    return call_number unless call_number && %w[LC DEWEY].include?(call_number_type) # Normalization only applied to LC/Dewey
```
jcoyne, line 153: The call_number_type is set to LC or DEWEY.  That would be not be possible with a nil call_number.

lib/sirsi_holding.rb
```diff
@@ -24,7 +24,7 @@ class SirsiHolding
   attr_reader :id, :current_location, :home_location, :library, :scheme, :type, :barcode, :public_note, :tag, :course_reserves
 
   # rubocop:disable Metrics/ParameterLists
-  def initialize(tag:, call_number:, home_location:, library:, barcode:, scheme: nil, current_location: nil,
+  def initialize(call_number:, home_location:, library:, barcode:, tag: nil, scheme: nil, current_location: nil,
```
jcoyne, line 27: Why are we allowing tag to be nil now?  There is no place in the code where we don't provide this.

jcoyne (changes requested):

lib/sirsi_holding.rb
cbeer: Again, am I missing something? The `call_number_type` method I'm looking at looks only at the scheme, and has nothing to do with the value of `call_number`. At least, in testing, I'm able to blow it up with:

`described_class.new(call_number: nil, scheme: 'LC', library: nil, home_location: nil, barcode: nil)`


```
  1) SirsiHolding#call_number with a nil call number returns an empty string
     Failure/Error: call_number = call_number.strip.gsub(/\s\s+/, ' ') # reduce multiple whitespace chars to a single space
     
     NoMethodError:
       undefined method `strip' for nil:NilClass
     # ./lib/sirsi_holding.rb:155:in `normalize_call_number'
     # ./lib/sirsi_holding.rb:44:in `call_number'
```

lib/sirsi_holding.rb
cbeer: I don't understand what `tag` is supposed to be used for, the test in #1057 didn't provide it, and we're not doing anything with it. That makes it seem pretty optional to me?

lib/sirsi_holding.rb
cbeer: I don't see why not? I don't think anything validates whether a call number is what you say it is.

thatbudakguy (approved):
going to merge this, since I think we're removing `tag` anyway

PR #1062, opened 2023-08-08
Remove SirsiHolding#tag

jcoyne:
We can use the other fields in SirsiHolding instead


thatbudakguy (approved):
I never figured out what this was supposed to be for anyway...

PR #1063, opened 2023-08-08
Parse identifiers when deriving MARC from FOLIO instance records

### LINKED ISSUE #1047: Enhance FOLIO instance -> MARC mapping to include identifiers
We want to map FOLIO instance identifiers to some MARC fields. We're hoping we can look at the format of the identifier string and map it into appropriate fields:

- [x] LCCNs `010a`: `%r{^(([ a-z]{3}\d{8})|([ a-z]{2}\d{10})) ?|( /.*)?$}`
- [x] ISBNs `020a`: `/^\d{9}[\dX].*/`, `/^(978|979)\d{9}[\dX].*/`, or `/^\d{12}[\dX].*/`
- [x] ISSNs `022a`: `/^\d{4}-\d{3}[X\d]\D*$/`
- [x] OCLC numbers `035a`: starts with `(OCoLC-M)` or `(OCoLC)`

(We're hoping to avoid resolving the identifierTypeId into the actual types, unless and until we actually need that data for something else.. our indexing for the fields is entirely pattern based so everything should map as expected without looking at the type names)

thatbudakguy:
Closes #1047

### DIFF
--- a/lib/folio_record.rb
+++ b/lib/folio_record.rb
@@ -273,12 +273,33 @@ class FolioRecord
     end
   end
 
-  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
   def instance_derived_marc_record
     MARC::Record.new.tap do |marc|
       marc.append(MARC::ControlField.new('001', hrid))
       # mode of issuance
       # identifiers
+      record.dig('instance', 'identifiers').each do |identifier|
+        id_value = identifier['value']
+
+        case id_value
+        # LCCN
+        when /^([ a-z]{3}\d{8} |[ a-z]{2}\d{10})/
+          field = MARC::DataField.new('010', ' ', ' ', ['a', id_value])
+        # ISBN
+        when /^\d{9}[\dX].*/, /^\d{12}[\dX].*/
+          field = MARC::DataField.new('020', ' ', ' ', ['a', id_value])
+        # ISSN
+        when /^\d{4}-\d{3}[X\d]\D*$/
+          field = MARC::DataField.new('022', ' ', ' ', ['a', id_value])
+        # OCLC
+        when /^\(OCoLC.*/
+          field = MARC::DataField.new('035', ' ', ' ', ['a', id_value])
+        end
+
+        marc.append(field) if field
+      end
+
       record.dig('instance', 'languages').each do |l|
         marc.append(MARC::DataField.new('041', ' ', ' ', ['a', l]))
       end
@@ -341,7 +362,7 @@ class FolioRecord
       # date updated
     end.to_hash
   end
-  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
   def folio_electronic_access_marc_field(eresource)
     ind2 = case eresource['name']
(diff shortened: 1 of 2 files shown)


lib/folio_record.rb
```diff
@@ -260,12 +262,33 @@ def stripped_marc_json
...
     MARC::Record.new.tap do |marc|
       marc.append(MARC::ControlField.new('001', hrid))
       # mode of issuance
       # identifiers
+      record.dig('instance', 'identifiers').each do |identifier|
+        id_value = identifier['value']
+
+        case id_value
+        # LCCN
+        when /^([ a-z]{0,3}\d{8} |[ a-z]{0,2}\d{10})$/
```
cbeer, line 276: I just noted on the ticket, we have real-world 010 values like
   68038902 //r84

that won't be caught by this revised regex.

jcoyne: It would be great if we could extract this to it's own class (e.g. `IdentifierClassifier` maybe?).  That would make it much easier to write tests for.

thatbudakguy: Should the extra data at the end of LCCNs be preserved as the identifier value? I.e. keep the entire ` 68038902 //r84` and not just ` 68038902`

spec/lib/folio_record_spec.rb
```diff
@@ -61,6 +61,65 @@
...
+          'instance' => {
+            'id' => '0e050e3f-b160-5f5d-9fdb-2d49305fbb0d',
+            'title' => 'The title',
+            'identifiers' => [
+              { 'value' => 'garbage' },
+              { 'value' => '(OCoLC-M)948533645' },
+              { 'value' => '2021236856' },
+              { 'value' => '2381-5868' },
+              { 'value' => '1485631076 (paperback)' },
+              { 'value' => '9781485631071 (paperback)' }
```
thatbudakguy, line 77: it looked to me from the LOC docs that the extra info in here like ` (paperback) ` often ends up in subfield `q`. should we try to separate it that way, so that subfield `a` is only the numeric content?

thatbudakguy: (question for @cbeer, I guess?)

cbeer: The indexer downstream is extracting the data:

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/sirsi_config.rb#L1656

So I'm not sure we care. If it we me, I wouldn't try too hard for these stub records 🤷‍♂️

lib/folio_record.rb
thatbudakguy: Perfect, that keeps it simple

thatbudakguy: Updated to remove the terminator and require (vs. just allow) the blank/alpha prefixes to avoid confusing with ISBNs.

lib/folio_record.rb
```diff
@@ -3,6 +3,8 @@
 require 'active_support/core_ext/module/delegation'
 require 'active_support/core_ext/enumerable'
 
+require 'debug'
+
```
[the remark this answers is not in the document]
thatbudakguy: Removed

lib/folio_record.rb
thatbudakguy: @jcoyne where would you want that to live?

cbeer: I suggest we handle extracting classes in a separate PR. Who knows, we might discover we need to pull the identifier type info anyway 🤷‍♂️

PR #1065, opened 2023-08-09
Extract classes from FolioRecord

jcoyne:
This helps keep FolioRecord from becoming a junk-drawer.


lib/folio/marc_record_instance_mapper.rb
```diff
@@ -0,0 +1,81 @@
...
+  # Creates a Marc Record for an Folio instance
+  class MarcRecordInstanceMapper
+    # rubocop:disable Metrics/AbcSize
+    def self.build(instance, holdings)
+      MARC::Record.new.tap do |marc|
+        marc.append(MARC::ControlField.new('001', instance['hrid']))
+        # mode of issuance
+        # identifiers
+        instance['identifiers'].each do |identifier|
+          field = MarcIdentifierMapper.for(identifier['value'])
```
cbeer, line 13: I'm fine with splitting all this mapping out from `FolioRecord` into `MarcRecordInstanceMapper`, but I think it'd be behoove us to keep all the MARC mapping logic contained and implemented as cleanly and understandably as possible so it is easier for our colleagues throughout SUL to grok. Split out `MarcIdentifierMapper` and `MarcElectronicAccess` just makes things way more complicated (for uncertain benefit? it's not like we have other consumers for this logic sitting aroubnd)

jcoyne: I think it makes it easier, because you can see that this method depends on no other methods. It's also more testable as you can write a test just about the identifier mapper without having to do the whole mapping.

cbeer: Why are we talking about 400 lines? It looks like re-incorporating those classes would add 40 lines, and would put all the FOLIO -> MARC mapping code together.

Again, I think that's a huge benefit to our colleagues, especially in metadata.

cbeer (changes requested):
See inline comment thread about extracted classes from `MarcRecordInstanceMapper`.

lib/folio/marc_record_instance_mapper.rb
jcoyne: I thought we're talking about putting it all in one class.  `MarcElectronicAccess` is used in both `MarcRecordMapper` and `MarcInstanceMapper`

cbeer: Here's what I'm talking about: https://github.com/sul-dlss/searchworks_traject_indexer/pull/1066

I don't think the duplication is a big deal. I would be shocked if these mappings changed with any kind of regularity.

jcoyne: Okay, can you make that a "ready" PR and I'll merge it?

PR #1067, opened 2023-08-11
Create a library store

jcoyne:


lib/folio/library_store.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Folio
+  # A cache of library data
+  class LibraryStore
```
cbeer, line 5: Can we just call this `Folio::Library` to match all our other code bases?

jcoyne: Sure, done.

Issue #1074, opened 2023-08-15
Where did some libraries go after the last full index?


cbeer:
Hoover, Media/microtext, and Rumsey all  used to show up in the Library facet with lots of stuff.

Possibly related to #993?

![Image](https://user-images.githubusercontent.com/111218/260743952-72d457dc-8b73-4d88-8325-3f7b770cdeb7.png)

E.g. https://searchworks-folio-dev.stanford.edu/view/622510 has no `building_facet` value

jcoyne:
It looks like this query is returning no results. Catkey is for https://searchworks.stanford.edu/view/14741648

```sql
      SELECT
        vi.id,
          jsonb_build_object(
            'instance',
              vi.jsonb || jsonb_build_object(
                'suppressFromDiscovery', COALESCE((vi.jsonb ->> 'discoverySuppress')::bool, false),
                'electronicAccess', COALESCE(sul_mod_inventory_storage.getElectronicAccessName(COALESCE(vi.jsonb #> '{electronicAccess}', '[]'::jsonb)), '[]'::jsonb),
                'statisticalCodes', CASE WHEN '4bc78766-8f34-4b1a-9e39-2a689a4ae998' = ANY(ARRAY(SELECT jsonb_array_elements_text(vi.jsonb->'statisticalCodeIds'))) THEN
                                        jsonb_build_array(
                                          jsonb_build_object(
                                              'id', '4bc78766-8f34-4b1a-9e39-2a689a4ae998',
                                              'name', 'Database'
                                          )
                                        )
                                    ELSE
                                       '[]'::jsonb
                                    END,
                'administrativeNotes', '[]'::jsonb,
                'notes', COALESCE((SELECT jsonb_agg(e) FROM jsonb_array_elements(vi.jsonb -> 'notes') AS e WHERE NOT COALESCE((e ->> 'staffOnly')::bool, false)), '[]'::jsonb)

              ),
            'source_record', COALESCE(jsonb_agg(DISTINCT mr."content"), '[]'::jsonb),
            'items',
              COALESCE(
                jsonb_agg(
                  DISTINCT item.jsonb || jsonb_build_object(
                    'suppressFromDiscovery',
                    CASE WHEN item.id IS NOT NULL THEN
                      COALESCE((vi.jsonb ->> 'discoverySuppress')::bool, false) OR
                      COALESCE((hr.jsonb ->> 'discoverySuppress')::bool, false) OR
[...]

PR #1078, opened 2023-08-15
Update FolioPostgresReader to accept 'addl_from' clauses for filtering

cbeer:
```
usage: script/process_folio_postgres_to_kafka.rb [options]
    --help           
    --kafka-topic    The kafka topic used for writing records
    --verbose        
    --processes      Number of parallel processes to spawn to handle querying

State management
    --state-file     
    --no-state-file  do not use the state file to track the current query time
    --full           enable full indexing (vs delta indexing, that uses the modified timestamp from the state file)

SQL query options
    --sql-query      a list of additional SQL filters to apply to the query
    --sql-join       an additional SQL join query to apply to the underlying query
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,3 +41,4 @@ group :deployment do
 end
 
 gem 'activesupport', '~> 7.0'
+gem 'slop'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -296,6 +296,7 @@ DEPENDENCIES
   rubocop
   ruby-kafka
   simplecov
+  slop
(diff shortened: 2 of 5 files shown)


jcoyne:
Can you explain a bit about what this is used for? When might we want to do `postgres.addl_from`?

cbeer:
I've pushed up part 2 of this change. We might want to use this to, say, reindex Hoover, Rumsey, and ARS (#1074) to get the right data without having to reindex all 10 million records.

PR #1079, opened 2023-08-15
Exclude some mapped FOLIO locations from being translated to Symphony codes

### LINKED ISSUE #1068: current_location field in `item_display_struct` should be the FOLIO code if we can't map it to a Symphony code for parity
We're currently lopping the library prefix off (somehow?), which breaks the downstream location mapping code for locations without a symphony equivalent (e.g. `SUL-TS-CC-REPAIR` should display in Searchworks as "Out for repair" instead of "TS-CC-REPAIR")

https://searchworks-folio-dev.stanford.edu/view/12278435
Image

corylown:
Fixes #1068

### DIFF
--- a/lib/locations_map.rb
+++ b/lib/locations_map.rb
@@ -34,7 +34,7 @@ class LocationsMap
               col_sep: "\t").each_with_object({}) do |(home_location, library_code, folio_code), hash|
       library_code = { 'LANE' => 'LANE-MED' }.fetch(library_code, library_code)
 
-      # Skipping location codes that we want handle as FOLIO codes in SearchWorks.
+      # Skipping location codes that we want to handle as FOLIO codes in SearchWorks.
       next if LOCATIONS_TO_SKIP.include?(folio_code)
 
       # SAL3's CDL/ONORDER/INPROCESS locations are all mapped so SAL3-STACKS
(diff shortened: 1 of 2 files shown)


lib/locations_map.rb
```diff
@@ -24,6 +34,9 @@ def load_home_locations
               col_sep: "\t").each_with_object({}) do |(home_location, library_code, folio_code), hash|
       library_code = { 'LANE' => 'LANE-MED' }.fetch(library_code, library_code)
 
+      # Skipping location codes that we want handle as FOLIO codes in SearchWorks.
```
jcoyne, line 37: typo here I think.

lib/locations_map.rb
```diff
@@ -37,6 +50,9 @@ def load_home_locations
   def load_current_locations
     CSV.parse(File.read(File.join(__dir__, 'translation_maps', 'temp_locations.tsv')),
               col_sep: "\t").each_with_object({}) do |(current_location, library_code, folio_code), hash|
+      # Skipping location codes that we want handle as FOLIO codes in SearchWorks.
```
jcoyne, line 53: here as well

cbeer, line 53:
```suggestion
      # Skipping location codes that we want to handle as FOLIO codes in SearchWorks.
```

lib/locations_map.rb
corylown: yes, thank you. Fixed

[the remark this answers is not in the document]
corylown: Fixed. thank you

PR #1083, opened 2023-08-15
Add support for indexing from a list of FOLIO ids

cbeer:
This might be faster than waiting for FOLIO's queries for some smaller collections. I'm not sure it's _that_ much faster in practice, but seems handy to have available 🤷‍♂️ 

The FOLIO UI's "download identifiers" provides them as quoted strings (reusing some CSV escaping code??), so I've accounted for that too.


jcoyne:
Need to rebase to resolve conflicts.

jcoyne (approved):

PR #1088, opened 2023-08-17
Only show the special CREZ loan period if it is different than normal

cbeer:
E.g. this is kinda funny:
Screenshot 2023-08-17 at 11 11 01


thatbudakguy (approved):
it almost seems like an error that something would have the same temp and permanent loan type, but this seems the simplest way to handle it i guess

PR #1091, opened 2023-08-17
Use location/details/searchworksGovDocsClassification to map govdocs …

### LINKED ISSUE #1075: Use location/details information (`searchworksGovDocsClassification`) to map govdocs locations for `callnum_facet_hsim`
- [ ] update `folio-graphql` to expect a string property called `searchworksGovDocsClassification` on `Location.details`
- [ ] set the `searchworksGovDocsClassification` value for gov doc locations in FOLIO according to the translation map (see below)
- [ ] update the logic for `SirsiHolding#gov_doc_loc?` to use the location details instead of `GOV_DOC_LOCS`
- [ ] update the indexer to use the value of `searchworksGovDocsClassification` in the call number facet
- [ ] extract the handling of marc 086 values to its own separate `to_field` definition (optional)

jcoyne:
…locations

Fixes #1075

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -234,6 +234,16 @@ to_field 'format_main_ssim' do |record, accumulator, _context|
   accumulator << 'Database' if record.statistical_codes.any? { |stat_code| stat_code['name'] == 'Database' }
 end
 
+to_field 'callnum_facet_hsim' do |record, accumulator, _context|
+  gov_doc_values = record.items.filter_map do |item|
+    item.dig('location', 'effectiveLocation', 'details', 'searchworksGovDocsClassification')
+  end
+
+  gov_doc_values.uniq.each do |gov_doc_value|
+    accumulator << ['Government Document', gov_doc_value].join('|')
+  end
+end
+
 ##
 # Skip records for missing `item_display` field
 each_record do |_record, context|


lib/folio_record.rb
```diff
@@ -49,6 +49,8 @@ def sirsi_holdings
       holding = holdings.find { |holding| holding['id'] == item['holdingsRecordId'] }
       next unless holding
 
+      gov_docs_classification = item.dig('location', 'effectiveLocation', 'details', 'searchworksGovDocsClassification')
```
cbeer, line 52: I'm surprised we have to snake this data through the `SirsiHolding` at all. It seems to me we could just pluck them out of `record.items` in the traject config.

lib/traject/config/sirsi_config.rb
```diff
@@ -1927,29 +1927,28 @@ def holdings(record, context)
...
-                              'Federal'
-                            else
-                              raw_location
-                            end
-        end
-      else
-        gov_doc_values << raw_location
+    if holding.gov_docs_classification
+      gov_doc_values << holding.gov_docs_classification
+    elsif marc_086.any?
```
cbeer, line 1939: Did this logic change? It seems like we dropped the outside condition?

```suggestion
    elsif raw_location == 'Other' && marc_086.any?
```

lib/traject/config/sirsi_config.rb
```diff
@@ -1927,29 +1927,28 @@ def holdings(record, context)
...
-                              'International'
-                            elsif marc_field.indicator1 == '0'
-                              'Federal'
-                            else
-                              raw_location
-                            end
-        end
-      else
-        gov_doc_values << raw_location
+    if holding.gov_docs_classification
```
cbeer, line 1937: Instead of mixing new FOLIO logic into this `to_field`, it might be cleaner to do it separately so it's easier to junk the old stuff?

cbeer:
👍 ; I also double-checked `SirsiHolding#skipped?` to make sure that short-circuit in the symphony code is unnecessary (and it should be... it's checking libraries and locations that can't exist if we're in one of those govdocs locations)

Issue #1094, opened 2023-08-18
Consolidate sirsi_config + folio_config indexers


cbeer:
The `sirsi_config` commit history has great value in deciphering previous choices. Our approach to consolidation should attempt to preserve that history (perhaps renaming `sirsi_config` to something sufficiently generic, e.g. `marc_config`, and integrating the `folio_config` into the renamed file?)

Before consolidating the actual code, we either need to revisit tests that use `marc_config` or some test harness wiring to work with `folio_config`/`Folio::Record`.

We can also replace the Symphony-specific `to_field` stanzas with the overrides from `folio_config`

PR #1101, opened 2023-08-21
Add a shim for stubbing properties needed by e.g. the folio indexer given a raw MARC record

cbeer:
Not pretty, but easy to automate until we touch each test manually.

Part of https://github.com/sul-dlss/searchworks_traject_indexer/issues/1094


jcoyne (approved):

lib/traject/config/folio_config.rb
```diff
@@ -251,7 +252,7 @@ def holdings(record, context)
 ##
 # Skip records for missing `item_display` field
 each_record do |_record, context|
-  if context.output_hash['item_display'].blank? && context.output_hash['url_fulltext'].blank?
+  if context.output_hash['item_display'].blank? && context.output_hash['url_fulltext'].blank? && settings['skip_empty_item_display'] > -2
```
jcoyne, line 255: Is there a reason we're testing on -2 here?  The docs says we'll skip on anything `> -1` https://github.com/sul-dlss/searchworks_traject_indexer/blob/0a4bf72d618f9fd8a28760ae97ab64bdd1af9f99/README.md?plain=1#L170

cbeer: Just for conveniently reviewing this PR separate from #1103 (where it changes back to `-1`: https://github.com/sul-dlss/searchworks_traject_indexer/pull/1103/files#diff-93251714c685d86f273f6f69bd7e74da3770b4c810dd1fc5f1a0720033586578R2361.)

We have two similar `skip` steps in `folio_config` and `marc_reader`, and I'm using -2 to skip the `marc_reader` version but not this one.

PR #1115, opened 2023-08-22
Don't skip records just because it has no unsuppressed items; some of…

cbeer:
… these are also e-resources.

We're checking if there are no items or full-text URLs after we've processed the record; there's no need to double-up on that check (although someday it'd be nice to check if the record will be indexed before we do all the hard work...).

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -120,11 +120,6 @@ each_record do |record, context|
   end
 end
 
-# Skip records that only have suppressed items
-each_record do |record, context|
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
Can we remove https://github.com/sul-dlss/searchworks_traject_indexer/blob/6f02c40820b3ea55bb230bb7587e4ce712ba5203/lib/folio_record.rb#L176-L178 too?

thatbudakguy:
#1117 supersedes this (because @cbeer asked me to add it while we were pairing on that work)

Issue #1116, opened 2023-08-22
Why does our index from folio-prod show so many on-order records?


thatbudakguy:
For example records marked on-order in sw-folio but not in sw-symphony:
- a11589740
- a14903045
- a14639313

thatbudakguy:
For most of these, it seemed like the holdings were being suppressed because they were in `-MIGRATE-ERR` locations, even though the items were correct and also had statuses like in-process. It seems like the solution is to allow `-MIGRATE-ERR`, but ask libsys to suppress from discovery in FOLIO instead.

Issue #1030, opened 2023-07-28
Don't index mholds with *MIGRATE-ERR


jcoyne:
https://folio-test.stanford.edu/inventory/view/532dee00-4455-555f-9ef2-6cc3f575a0b1/e801da87-8e78-5ea9-aa9a-9b25e80c02f5?qindex=hrid&query=a3357319&sort=title
https://searchworks-folio-dev.stanford.edu/view/3357319

This causes this panel to display:
<img width="368" alt="Screenshot 2023-07-28 at 12 38 39 PM" src="https://github.com/sul-dlss/searchworks_traject_indexer/assets/92044/32337295-8448-4c12-b0b9-ad3ba4342d37">

PR #1117, opened 2023-08-22
Allow indexing of holdings and items in MIGRATE-ERR locations

thatbudakguy:
- Don't filter out holdings and items in *-MIGRATE-ERR locations
- Remove check for skipping indexing of records with all suppressed items

### DIFF
--- a/lib/folio/mhld_builder.rb
+++ b/lib/folio/mhld_builder.rb
@@ -28,8 +28,7 @@ module Folio
 
     private
 
-    # Remove suppressed records, electronic records, records that have a location matching *-MIGRATE-ERR and
-    # records with no holdings statement
+    # Remove suppressed records, electronic records, and records with no holdings statement
     def filtered_holdings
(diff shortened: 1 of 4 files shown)


jcoyne:
@thatbudakguy Can you say more about why are we doing this work?

thatbudakguy:
See the comment on #1116

jcoyne:
@thatbudakguy won't this cause #1030 to reappear?

thatbudakguy:
I think so, yes — we need them to be suppressed from discovery in order to resolve that.

jcoyne:
That sounds like a lot of work. Should we hold this until those data changes have been made?  The second part of this change duplicates work Chris did in #1115.  Can you revert that?

cbeer:
I've closed my PR. I think it's appropriate to adjust all the suppressed settings here.

We don't need to wait for the data work; libsys will be fixing it, and we'll be much happier diagnosing problems if we're not missing pieces of data.

PR #1118, opened 2023-08-22
Unpend tests for govdocs callnum

### LINKED ISSUE #1110: Consolidate govdoc call number tests
The existing tests check for old behavior using the `999` field:

https://github.com/sul-dlss/searchworks_traject_indexer/blob/298a1e276f1096f2ad8f6885f49e8a710661eff0/spec/lib/traject/config/callnum_facet_spec.rb#L568-L636

They need to be updated to use the location details and cover the implementation.

jcoyne:
Fixes #1110


spec/integration/folio_config_spec.rb
```diff
@@ -898,76 +898,4 @@
       it { is_expected.to eq ['Music score'] }
     end
   end
-
-  describe 'callnum_facet_hsim' do
```
jcoyne, line 902: I moved this test into callnum_facet_spec.rb

corylown (approved):

PR #1127, opened 2023-08-23
Remove duplicate or unnecessary callnum_facet tests

jcoyne:
There is no code for call num types other than Dewey or LC, so we don't need to explicitly test every combination of that. We also don't need to test invalid LC or Dewey with any other context as we already have tests for that

### DIFF
--- a/spec/lib/traject/config/callnum_facet_spec.rb
+++ b/spec/lib/traject/config/callnum_facet_spec.rb
@@ -43,16 +43,6 @@ RSpec.describe 'Call Number Facet' do
   end
 
   describe 'call numbers excluded for various reasons' do
-    it 'handles unexpected callnum type (by not including them)' do
-      expect(record_with_holdings(call_number: 'M123 .M234', scheme: 'ALPHANUM', indexer:)[field]).to be_nil
-      expect(record_with_holdings(call_number: 'M123 .M234', scheme: 'HARVYENCH', indexer:)[field]).to be_nil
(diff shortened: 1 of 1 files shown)


spec/lib/traject/config/callnum_facet_spec.rb
```diff
@@ -325,13 +260,9 @@ def record_with_999(call_number:, scheme:, indexer:, home_location: 'STACKS', li
         # Harvard Yenching call numbers
         '6.4C-CZ[BC]',
         '2345 5861 V.3',
-        '2061 4246 NO.5-6 1936-1937',
```
jcoyne, line 328: Removed these 4 because they had schemes that we don't process into callnum_facet_hsim

spec/lib/traject/config/callnum_facet_spec.rb
```diff
@@ -157,37 +117,6 @@ def record_with_holdings(call_number:, scheme:, indexer:, home_location: 'STACKS
...
-    end
-
-    it 'does not return call numbers w/ the scheme ASIS' do
-      expect(record_with_holdings(call_number: '(ADL4044.1)XX', scheme: 'ASIS', indexer:)[field]).to be_nil
-      expect(record_with_holdings(call_number: '134776', scheme: 'ASIS', indexer:)[field]).to be_nil
-      expect(record_with_holdings(call_number: 'INTERNET RESOURCE', scheme: 'ASIS', indexer:)[field]).to be_nil
-    end
-  end
-
-  describe 'LC Call Numbers' do
```
marlo-longley, line 173: These should still be included, we think?

marlo-longley (approved):

PR #1133, opened 2023-08-23
Prune skipped locs

jcoyne:
Done in two steps:
* Remove any skipped locations that was not mapped to folio
* Remove SKIPPED_LOCS that are filtered out in LocationsMap


spec/lib/traject/config/callnum_facet_spec.rb
```diff
@@ -60,12 +60,6 @@ def record_with_999(call_number:, scheme:, indexer:, home_location: 'STACKS', li
       expect(record_with_999(call_number: 'M123 .M234', scheme: 'AUTO', indexer:)[field]).to be_nil
     end
 
-    it 'handles skipped items (by not including them)' do
```
marlo-longley, line 63: Do we want to fully delete this test or just replace the loc with one of the included ones, like `SHADOW` ?

jcoyne: Yeah, good idea. I've added a test for SHADOW.

Issue #1109, opened 2023-08-22
Consolidate `date_cataloged` tests


cbeer:
We're no longer using the `916b` to populate `date_cataloged`. Make sure there's no value in the older tests, and then... throw them out? 

https://github.com/sul-dlss/searchworks_traject_indexer/blob/298a1e276f1096f2ad8f6885f49e8a710661eff0/spec/integration/folio_config_spec.rb#L80-L110

https://github.com/sul-dlss/searchworks_traject_indexer/blob/298a1e276f1096f2ad8f6885f49e8a710661eff0/spec/lib/traject/config/date_spec.rb#L543-L567

PR #1135, opened 2023-08-23
Removing 916b for date cataloged, slightly updating folio config spec for date cataloged

hudajkhan:
Related to #1109.

What this pull request does:
a) Remove the date cataloged portion in the date_spec tests, since we are no longer relying on 916b to get the cataloged date.  The associated fixture file, "newItemsDataCataloged.xml", is also no longer required with the removal of those test conditions. 
b) The indexing logic for the cataloged date from Folio is here:  https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/folio_config.rb#L885 .  I added a condition to check where the parsing of the date would throw an error.   If an error is thrown, we expect that the value is not indexed/that field is empty. I also updated the wording of some of the contexts.  

I considered merging the folio config date cataloged test into date_spec.  That would require incorporating folio client, folio record, etc. and these are not used anywhere else in date_spec.   If I should focus on that, I can move the folio_config_spec date cataloged tests into date_spec.


marlo-longley:
Nice @hudajkhan  -- I am wondering if we still need to be testing for ISO-ness in the date_spec. Is there a way to mock the FOLIO data rather than using the deleted XML?

hudajkhan:
@marlo-longley I think you're referring to the date_spec iso check that has been removed, and whether it would be good to maintain a check for iso in the folio config. I'll review that and see what it requires. Thanks!

hudajkhan:
@marlo-longley does this cover the iso case? I don't think that was visible in the changes view b/c that test was already there in folio_config

marlo-longley:
@hudajkhan ah thank you. I see it on line 106 of the folio_config_spec !

marlo-longley (approved):

PR #1141, opened 2023-08-24
Avoid testing with a 999 with ASIS as a scheme

jcoyne:
This can never occur in Folio

### DIFF
--- a/spec/lib/traject/config/format_main_spec.rb
+++ b/spec/lib/traject/config/format_main_spec.rb
@@ -242,19 +242,6 @@ RSpec.describe 'Format main config' do
     end
   end
 
-  context 'online copy only of a Software/Multimedia and database' do
-    let(:record) do
-      MARC::Record.new.tap do |r|
-        r.leader = '02441cms a2200517 a 4500'
-        r.append(MARC::ControlField.new('008', '920901d19912002pauuu1n    m  0   a0eng  '))
-      end
-    end
-
-    xit 'is only a database' do
-      expect(result[field]).to eq ['Database']
-    end
-  end
-
   context 'both physical copy and online copy of a Software/Multimedia and database' do
     let(:folio_record) do
       FolioRecord.new({
--- a/spec/lib/traject/config/format_physical_spec.rb
+++ b/spec/lib/traject/config/format_physical_spec.rb
@@ -347,7 +347,7 @@ RSpec.describe 'Format physical config' do
     end
   end
 
-  context 'based on 8833535' do
+  context 'when acccess_facet is "At the library" (based on 8833535)' do
     let(:record) do
       MARC::Record.new.tap do |r|
         r.leader = '02229cjm a2200409Ia 4500'
@@ -366,21 +366,15 @@ RSpec.describe 'Format physical config' do
     end
   end
 
-  context 'based on 8833535, but online' do
+  context 'when acccess_facet is not "At the library" (based on 8833535)' do
     let(:record) do
       MARC::Record.new.tap do |r|
         r.leader = '02229cjm a2200409Ia 4500'
         r.append(MARC::ControlField.new('007', 'sd fungnnmmneu'))
-        r.append(MARC::DataField.new('999', ' ', ' ',
-                                     MARC::Subfield.new('a', 'INTERNET RESOURCE'),
-                                     MARC::Subfield.new('w', 'ASIS'),
-                                     MARC::Subfield.new('i', '2475606-5001'),
-                                     MARC::Subfield.new('l', 'INTERNET'),
-                                     MARC::Subfield.new('m', 'SUL')))
       end
     end
 
-    it 'is not a CD, since it is online-only' do
+    it 'is nil' do
       expect(result[field]).to eq nil
     end
   end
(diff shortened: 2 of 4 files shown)


spec/lib/traject/config/format_main_spec.rb
```diff
@@ -236,38 +236,11 @@
...
-                                     MARC::Subfield.new('a', 'INTERNET RESOURCE'),
-                                     MARC::Subfield.new('w', 'ASIS'),
-                                     MARC::Subfield.new('i', '2475606-5001'),
-                                     MARC::Subfield.new('l', 'INTERNET'),
-                                     MARC::Subfield.new('m', 'SUL'),
-                                     MARC::Subfield.new('t', 'DATABASE')))
-      end
-    end
-
-    xit 'is only a database' do
```
jcoyne, line 254: This has been skipped for 5 years, so tossing it out.

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -236,38 +236,11 @@
...
-  end
-
   context 'both physical copy and online copy of a Software/Multimedia and database' do
     let(:record) do
       MARC::Record.new.tap do |r|
         r.leader = '02441cms a2200517 a 4500'
         r.append(MARC::ControlField.new('008', '920901d19912002pauuu1n    m  0   a0eng  '))
-        r.append(MARC::DataField.new('999', ' ', ' ',
-                                     MARC::Subfield.new('a', 'INTERNET RESOURCE'),
-                                     MARC::Subfield.new('w', 'ASIS'),
```
marlo-longley, line 266: Make this LC or Dewey instead, and keep the INTERNET RESOURCE part

spec/lib/traject/config/format_physical_spec.rb
```diff
@@ -370,12 +370,6 @@
       MARC::Record.new.tap do |r|
         r.leader = '02229cjm a2200409Ia 4500'
         r.append(MARC::ControlField.new('007', 'sd fungnnmmneu'))
-        r.append(MARC::DataField.new('999', ' ', ' ',
```
marlo-longley, line 373: We could probably delete this whole test not just the 999

spec/lib/traject/config/format_physical_spec.rb
```diff
@@ -549,12 +543,6 @@
       MARC::Record.new.tap do |r|
         r.leader = '01002cjm a2200313Ma 4500'
         r.append(MARC::ControlField.new('007', 'sd dmsdnnmslne'))
-        r.append(MARC::DataField.new('999', ' ', ' ',
```
marlo-longley, line 552: Same as above (delete?)

spec/lib/traject/config/format_physical_spec.rb
```diff
@@ -587,12 +575,6 @@
       MARC::Record.new.tap do |r|
         r.leader = '02683cjm a2200565ua 4500'
         r.append(MARC::ControlField.new('007', 'sdubsmennmplue'))
-        r.append(MARC::DataField.new('999', ' ', ' ',
```
marlo-longley, line 590: same as above

spec/lib/traject/config/format_physical_spec.rb
```diff
@@ -753,12 +735,6 @@
       MARC::Record.new.tap do |r|
         r.leader = '01205cim a2200337Ia 4500'
         r.append(MARC::ControlField.new('007', 'ss lunjlc-----'))
-        r.append(MARC::DataField.new('999', ' ', ' ',
```
marlo-longley, line 756: same

spec/lib/traject/config/format_physical_spec.rb
jcoyne: I updated the context instead to show why it is relevant to the code.

spec/lib/traject/config/format_main_spec.rb
jcoyne: done

spec/lib/traject/config/format_physical_spec.rb
marlo-longley: @jcoyne I was looking into the following `it` condition here, the whole test being:

```
  context 'when acccess_facet is not "At the library" (based on 8833535)' do
    let(:record) do
      MARC::Record.new.tap do |r|
        r.leader = '02229cjm a2200409Ia 4500'
        r.append(MARC::ControlField.new('007', 'sd fungnnmmneu'))
      end
    end

    it 'is not a CD, since it is online-only' do
      expect(result[field]).to eq nil
    end
  end
```

I'm not sure the "since it is online-only' makes sense here. The access_facet is now just empty. Should it read "it is not a CD, since it is not at the library"? Does this even make sense?

jcoyne: Okay I've just updated to say what they are (nil) instead of what they are not.

marlo-longley: Nice that makes more sense to me!

marlo-longley (approved):

PR #1143, opened 2023-08-24
Reduce use of 999 field in format_main spec

jcoyne:
A lot of these tests just include 999 fields even though the code doesn't care about them.

### DIFF
--- a/spec/lib/traject/config/format_main_spec.rb
+++ b/spec/lib/traject/config/format_main_spec.rb
@@ -241,13 +241,6 @@ RSpec.describe 'Format main config' do
       MARC::Record.new.tap do |r|
         r.leader = '02441cms a2200517 a 4500'
         r.append(MARC::ControlField.new('008', '920901d19912002pauuu1n    m  0   a0eng  '))
-        r.append(MARC::DataField.new('999', ' ', ' ',
-                                     MARC::Subfield.new('a', 'INTERNET RESOURCE'),
-                                     MARC::Subfield.new('w', 'ASIS'),
-                                     MARC::Subfield.new('i', '2475606-5001'),
-                                     MARC::Subfield.new('l', 'INTERNET'),
-                                     MARC::Subfield.new('m', 'SUL'),
-                                     MARC::Subfield.new('t', 'DATABASE')))
       end
     end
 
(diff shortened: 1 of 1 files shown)


spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1167,36 +1142,11 @@
     end
   end
 
-  context 'format serial publication:  leader/07 s and 008/21 blank (ignore LCPER in 999w)' do
```
jcoyne, line 1170: This test is now a duplicate of the one below as 999w has no role in the code.

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1348,144 +1292,94 @@
...
-        r.append(MARC::DataField.new('999', ' ', ' ',
-                                     MARC::Subfield.new('m', 'LANE-MED')))
-      end
-    end
-
-    it 'is a Book and not Archive/Manuscript' do
-      expect(result[field]).to eq ['Book']
-    end
-  end
+  context 'when library is LANE-MED' do
```
jcoyne, line 1295: Here we move all the LANE-MED tests under a single context.

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1348,144 +1292,94 @@
...
+    context 'when leader/07 is c' do
+      let(:record) do
+        MARC::Record.new.tap do |r|
+          r.leader = '01952cac  2200457Ia 4500'
+        end
       end
-    end
 
-    it 'is has no format' do
-      expect(result[field]).to eq nil
```
jcoyne, line 1445: These no format tests were not actually testing anything in our code, so they've been discarded.

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1522,263 +1416,52 @@
...
+        end
       end
     end
-
-    it 'is a book' do
-      expect(result[field]).to eq ['Book']
-    end
-  end
-
-  context '999 ALPHANUM starting with MFICHE  but not SPEC-COLL' do
```
jcoyne, line 1752: There is no special handling for this case, so removing.

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1932,44 +1596,6 @@
     end
   end
 
-  context '999 ALPHANUM starting with MFLIM' do
```
jcoyne, line 1935: There is no special handling for these cases so removing.

hudajkhan (approved):

PR #1148, opened 2023-08-24
Extract the method for creating a Folio record without stubbing holdings

jcoyne:
This makes it easier to understand which tests depend on the 999 field


marlo-longley:
@jcoyne this needs a rebase. Looks good.

My takeaway is that with this PR there are only 17 references to `marc_to_folio_with_stubbed_holdings` -- these are the ones that will need the 999 references extracted. Makes the work much easier, nice!

Issue #1058, opened 2023-08-07
Revisit SirsiHolding#skipped?


jcoyne:
I think a number of the cases we are checking for will no longer exist.
https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/sirsi_holding.rb#L47-L52

PR #1177, opened 2023-08-31
Extract FolioHolding

cbeer:
I think the two functional changes are:

- bound-withs are no longer necessarily SEE-OTHER (which... turns out not to be true in Symphony for some `GREEN/FED-DOCS`). Searchworks is figuring out something is bound-with from the holdings type, so the location code can be whatever.
- barcodes are optional; I think we resolved anything downstream in searchworks that required a barcode?

Fixes https://github.com/sul-dlss/searchworks_traject_indexer/issues/1058

### DIFF
--- a/spec/lib/folio_record_spec.rb
+++ b/spec/lib/folio_record_spec.rb
@@ -313,6 +313,7 @@ RSpec.describe FolioRecord do
       context 'when the bound with child is in SAL3' do
         let(:folio_record) { described_class.new(JSON.parse(File.read(file_fixture('folio_bw_child_see-other.json'))), client) }
         it 'adds SEE-OTHER as the home_location' do
+          skip('Nah')
           expect(folio_record.sirsi_holdings.first.home_location).to eq('SEE-OTHER')
         end
       end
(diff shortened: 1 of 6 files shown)


jcoyne (approved):

spec/lib/folio_record_spec.rb
```diff
@@ -313,6 +313,7 @@
       context 'when the bound with child is in SAL3' do
         let(:folio_record) { described_class.new(JSON.parse(File.read(file_fixture('folio_bw_child_see-other.json'))), client) }
         it 'adds SEE-OTHER as the home_location' do
+          skip('Nah')
```
jcoyne, line 316: Should we delete this?

spec/lib/folio_record_spec.rb
```diff
@@ -345,8 +346,8 @@
         it 'creates a stub bound-with item' do
           expect(folio_record.sirsi_holdings.first).to have_attributes(
             id: nil,
-            barcode: '14154194-1001',
-            home_location: 'SEE-OTHER',
+            # barcode: '14154194-1001',
```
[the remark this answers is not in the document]
cbeer: I left them here so it is clear what we're losing, so someone else can (more) easily evaluate the impact.

spec/lib/folio_record_spec.rb
cbeer, line 316:
```suggestion
          skip('Unclear whether we need to preserve this behavior in FOLIO')
```

PR #1184, opened 2023-09-01
Test for the expected value

jcoyne:
It's just simpler to show what the value will be than what it contains


spec/integration/compare_against_solrmarc_docs_spec.rb
```diff
@@ -47,7 +47,7 @@
 
       it 'maps collection' do
         # FOLIO add 'folio' to the list
-        expect(result['collection']).to include(*expected_doc['collection'])
+        expect(result['collection']).to eq ['folio', 'sirsi']
```
cbeer, line 50: This is even more confusing to me. The whole point of this test suite is to compare what we're doing with our old implementations. A test for "what is the current behavior" should go in a different file.

It looks like we might have one in `managed_purl_spec`, but maybe we can write a set of more specific tests.

(Not to mention, the old solrmarc documents don't have a `folio` value)

PR #1186, opened 2023-09-01
Refactor spec to use JSON MARC rather than solrmarc output

jcoyne:


spec/integration/compare_against_solrmarc_docs_spec.rb
```diff
@@ -2,64 +2,31 @@
...
 
-  Dir.glob(File.expand_path('solrmarc_example_docs/*', file_fixture_path)).each do |fixture|
-    context "with #{fixture}" do
-      let(:file) { File.read(fixture) }
-      let(:data) { JSON.parse(file) }
-      let(:solrmarc_doc) { data['doc'] }
       let(:expected_doc) do
-        data['doc'].transform_values { |v| Array(v).map(&:to_s).sort }
+        result = nil
+        File.foreach(File.expand_path('output.jsonl', file_fixture_path)) do |line|
```
cbeer, line 19: I'm not sure I understand what's going on here, how we'd add new documents to those jsonl files if we discover other differences between the indexing approaches, or what we'd do when we decide to intentionally diverge from the previous implementation.

jcoyne: It's finding the line in the output file with the id that matches the input record. We can add/change/remove any pair of lines from the input and output file if we want to change the tests.

jcoyne:
Superseded by #1239

PR #1200, opened 2023-09-05
Refactor spec to be affirmative in what we're testing

jcoyne:
and to avoid duplication

### DIFF
--- a/spec/lib/traject/config/item_info_spec.rb
+++ b/spec/lib/traject/config/item_info_spec.rb
@@ -33,6 +33,7 @@ RSpec.describe 'ItemInfo config' do
     let(:records) { MARC::XMLReader.new(file_fixture(fixture_name).to_s).to_a }
     let(:fixture_name) { 'buildingTests.xml' }
     let(:field) { 'building_facet' }
+    subject(:value) { result[field] }
 
     it 'has data' do
       expect(select_by_id('229800')[field]).to eq ['Archive of Recorded Sound']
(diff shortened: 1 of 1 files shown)


spec/lib/traject/config/item_info_spec.rb
```diff
@@ -86,24 +87,9 @@
...
-        'SUL',
-        'PHYSICS', # Physics Library
-        'MEYER', # INDEX-168 Meyer
-        'BIOLOGY', # closed
-        'CHEMCHMENG', # closed
-        'MATH-CS' # closed
-      )
+      invalid_buildings = results.select { |item| item[field].nil? }
+      ids_for_invalid_buildings = invalid_buildings.map { |n| n['id'].first }
+      expect(ids_for_invalid_buildings).to eq %w[1111 2222 115472 460947 919006 1391080 1732616 2442876 2797607 3142611 3277173 6493823 7117119 10421123]
```
cbeer, line 92: I'm not sure this is a case that can happen any more, but I'm confused about what this rewritten test is testing. Previously, it was clear that those random buildings weren't present... and now, I'm not sure?

jcoyne: I was concerned that it was looking for the non-presence of some codes. However, all the other values in that array were decoded, human readable labels.  This seems like it probably wasn't a good test, so instead I'm testing for all of the record ids in the results  where the building_facet is nil.

PR #1204, opened 2023-09-05
Exclude subfield 7 from 1xx and 7xx values

cbeer:
fixes SW-4173


jcoyne:
Lane is putting the string "Stanford Related" in the $7, which causes this text to erroneously show up on https://searchworks.stanford.edu/view/L292609 
Screenshot 2023-09-05 at 1 32 19 PM

Issue #1189, opened 2023-09-01
Replace holdings tests that use stub MARC records with `999` fields


thatbudakguy:
We have a large number of tests that set up a item record using the MARC 999 field:

```ruby
MARC::DataField.new('999'
```

We should be able to replace these by stubbing `FolioRecord#sirsi_holdings` with `SirsiHoldings` instances.

- [x] access_spec
- [x] database_az_subject_spec
- [x] format_main_spec
- [x] format_physical_spec
- [x] location_facet_spec
- [x] preferred_barcode_spec
- [x] sirsi_config_spec
- [ ] item_info_spec

PR #1205, opened 2023-09-05
Stub holdings records instead of relying on MARC 999

jcoyne:
Ref #1189


spec/lib/traject/config/item_info_spec.rb
```diff
@@ -837,66 +782,6 @@
           expect(result['item_display_struct'].map { |x| JSON.parse(x) }.first['note']).to eq '.PuBlIc. Note'
         end
       end
-
-      context 'when the public note does not have periods around it' do
```
jcoyne, line 841: These are currently not doing anything as the helper method `marc_to_folio_with_stubbed_holdings` is removing these before they become part of the FolioRecord.

thatbudakguy (approved):

Issue #1093, opened 2023-08-18
Add separate item_display entries for temporary location + status (vs lumping them into current_location)


cbeer:
Also, consider adding a permanent location code (?) intended to replace library/location

PR #1210, opened 2023-09-05
Add location and status to the item_display_struct

jcoyne:
Ref #1093

### DIFF
--- a/lib/folio_record.rb
+++ b/lib/folio_record.rb
@@ -175,6 +175,8 @@ class FolioRecord
         call_number: [item.dig('callNumber', 'callNumber'), item['volume'], item['enumeration'], item['chronology']].compact.join(' '),
         current_location: (current_location unless current_location == home_location_code).presence,
         home_location: home_location_code,
+        location: item_location_code,
+        status: item['status'],
         library: library_code,
         scheme: call_number_type_map(item.dig('callNumberType', 'name') || item.dig('callNumber', 'typeName')),
         type: item['materialType'],
@@ -208,6 +210,8 @@ class FolioRecord
         scheme: call_number_type_map(holding.dig('callNumberType', 'name')),
         # parent item's barcode
         barcode: parent_item['barcode'] || "#{hrid.sub(/^a/, '')}-#{(all_holdings.index(holding) + 1).to_s.ljust(3, '0')}1",
+        location: item_location_code,
+        status: parent_item['status'],
         # parent item's current location or SEE-OTHER (SAL3)
         # For the SAL3 logic, see https://consul.stanford.edu/display/MD/Bound+withs
         # When the bound-with item is in SAL3, both the Home and Current Locations on the child records should always be SEE-OTHER.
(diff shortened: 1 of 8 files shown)


lib/traject/config/sirsi_config.rb
```diff
@@ -1803,6 +1803,8 @@ def holdings(record, context)
         call_number: (item['a'] || '').strip,
         current_location: item['k'],
         home_location: item['l'],
+        location: 'PLACEHOLDER',
+        status: 'Available',
```
cbeer, line 1807:
```suggestion
```

I'm not sure if you've added these because they're now required, but we shouldn't populate them with wrong data either.

spec/support/result_helpers.rb
```diff
@@ -11,6 +11,8 @@ def marc_to_folio_with_stubbed_holdings(marc_record, client: stub_folio_client,
             call_number: (item['a'] || '').strip,
             current_location: item['k'],
             home_location: item['l'],
+            status: 'Available',
+            location: 'GRE-STACKS',
```
cbeer, line 15: Will this "wrong" data confuse us later when we're trying to get away from the 999 data? I think it's better to leave it blank or map it from the data we do have.

lib/folio_record.rb
```diff
@@ -256,6 +263,8 @@ def on_order_stub_holdings
         call_number: nil,
         scheme: nil,
         library: lib,
+        status: 'On order',
+        location: 'GRE-STACKS',
```
cbeer, line 267: I'm not sure where `GRE-STACKS` came from here?

lib/traject/config/sirsi_config.rb
jcoyne: what would be appropriate values?

cbeer: `nil`?

lib/sirsi_holding.rb
```diff
@@ -16,15 +16,20 @@ class SirsiHolding
...
 
-  attr_reader :id, :current_location, :home_location, :library, :scheme, :type, :barcode, :public_note, :course_reserves
+  attr_reader :id, :current_location, :home_location, :location, :status, :library, :scheme, :type, :barcode, :public_note, :course_reserves
 
+  # @param [String] location the folio code for location
+  # @param [Status] status the folio status
+  # @param [String] current_location Legacy parameter to be replaced by "location" and "status"
   # rubocop:disable Metrics/ParameterLists
-  def initialize(call_number:, home_location:, library:, barcode:, scheme: nil, current_location: nil,
+  def initialize(call_number:, home_location:, library:, barcode:, location:, status:, scheme: nil, current_location: nil,
```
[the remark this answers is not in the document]
jcoyne: I've mostly been using the existing code that uses temporarylocation if it exists and searchworksTreatTemporaryLocationAsPermanentLocation is true, else permanent location otherwise effectiveLocation. This is the value that we've been using to map to home_location. I don't know what we want to call this. Is that not the value you want me to use?

jcoyne:
Superseded by #1241

PR #1220, opened 2023-09-07
Remove User attributes from the service definition

cbeer:
`User` is unnecessary (and possibly harmful?) for user-level systemd services.


jcoyne:
rubocop doesn't like this

PR #1236, opened 2023-09-11
Rewrite tests so that we don't need to use MARC 999

jcoyne:
These tests now stub the holdings directly, since Folio doesn't use MARC 999


jcoyne:
Superceded by #1238

PR #1245, opened 2023-09-14
Switch the holding factory to produce FolioHoldings

jcoyne:


spec/factories/holdings.rb
```diff
@@ -21,12 +21,12 @@
 
     factory :sudoc_holding do
       call_number { 'I 19.76:98-600-B' }
-      scheme { 'SUDOC' }
+      scheme { 'superintendent' }
```
cbeer, line 24: Should we use plausible FOLIO values (from https://github.com/sul-dlss/folio_migration/blob/63166e3163bcee0562dd0c9b96d313e1adf968ac/mapping_files/call_number_type_mapping.tsv#L6)?

```suggestion
      scheme { 'Superintendent of Documents classification' }
```

spec/factories/holdings.rb
```diff
@@ -21,12 +21,12 @@
...
     end
 
     factory :alphanum_holding do
       call_number { 'ISHII SPRING 2009' }
-      scheme { 'ALPHANUM' }
+      scheme { 'title' }
```
cbeer, line 29:
```suggestion
      scheme { 'Shelving control number' }
```

PR #1247, opened 2023-09-14
Index mhld_display as a json field

jcoyne:
This makes it easier to work with


lib/traject/config/folio_config.rb
```diff
@@ -2330,7 +2330,11 @@ def accumulate_summary_struct_fields(matching_fields, tag, label, marc, accumula
 to_field 'on_order_library_ssim', extract_marc('596a', translation_map: 'library_on_order_map')
 
 to_field 'mhld_display' do |record, accumulator, _context|
-  record.mhld.each { |holding| accumulator << holding }
+  record.mhld.each { |holding| accumulator << [holding.library, holding.location, holding.public_note, holding.library_has, holding.latest].join(' -|- ') }
+end
+
+to_field 'mhld_display_struct' do |record, accumulator, _context|
+  record.mhld.each { |holding| accumulator << holding.to_h }
```
cbeer, line 2337: If we're starting something new, do we really want/need to replicate the old structure here, or is it worth mirroring either the FOLIO data structures or how Searchworks actually uses this data?

Back in https://github.com/sul-dlss/searchworks_traject_indexer/pull/876/files, I suggested one possible structure.

jcoyne: @cbeer I added a second commit to this PR.  Is this more like what you have envisioned?

cbeer: If it were me (and it was, in #876), I'd go a little further. At the very least,  `latest` is always the same value for a location, so there's no reason to keep duplicating it for each holdings note.

In #876, you'll see I proposed a data structure that Searchworks should be happy to consume and requires less processing in the indexer:

```
library:
  location:
    holdings: []
    index: []
    supplements: []
    latest_received: ''
    note: ''
```

It's not clear to me how the FOLIO model changes that general approach, but:
- I think `note` could be multivalued?
- in MARC, the 866, 867, and 868 (e.g. holdings, index, supplements) could be freely interleaved and maybe order was important. In FOLIO, they're separate fields so I feel even better about breaking them apart as above.

Maybe there's other things; I'd encourage you to look at the FOLIO data model, what Searchworks is actually doing this this data, and the backlog of MHLD-adjacent tickets in order to figure out what we should do:
- https://github.com/sul-dlss/SearchWorks/issues/1974
- SW-3886
- SW-3824
- SW-1556
- etc

jcoyne: @cbeer I'm not sure I understand index or supplements. Are these somewhere in the database query?

jcoyne: @cbeer I'm not seeing where searchworks needs to differentiate between regular holding statements, index holding statements and suppplement holding statements.  Can you explain more about why you chose to take the data model in that direction?

cbeer: That is essentially FOLIO's data model. We currently differentiate between them here in the indexer:

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/folio/mhld_builder.rb#L67-L73

jcoyne: Sure, but Searchworks needs them merged together, so is there any reason not to just have one set?

PR #1255, opened 2023-09-20
Remove old Symphony-style course reserves indexing

cbeer:


lib/traject/config/folio_config.rb
```diff
@@ -2560,52 +2551,6 @@ def accumulate_summary_struct_fields(matching_fields, tag, label, marc, accumula
...
-      # This is not dissimilar to the original building_facet mapping:
-      # Try the current location first, in case it has an overridden library, and then
-      # fall back on the library code.
-      library_map[item_display_value[:current_location]] || library_map[item_display_value[:library]]
-    end
-  end
-
-  context.output_hash['building_facet'] = new_building_facet_vals.uniq if new_building_facet_vals.any?
-end
-
```
thatbudakguy, line 2608: Just for my understanding: are we allowed to remove these because we look for and use temp locations at the time we construct `item_display_struct` and `building_facet`, and thus have no need to revisit/alter them? What about stuff like `course_id` and `loan_period`? I don't see us adding that `item_display_struct` anywhere else.

cbeer: Right -- the FOLIO item record is actually right up-front.

thatbudakguy: Oh, I see — and `courses_json_struct` handles the rest, I guess? cool.

PR #1263, opened 2023-09-25
Rewrite MEDIA-CENTER to MEDIA-MTXT in LocationsMap

thatbudakguy:
This handles a case where the FOLIO temp location of MEDIA-CRES was
mapped to MEDIA-CENTER for its library, which wasn't a library
in Symphony. We're currently preserving the symphony library codes
to pass to SW, so we need to rewrite it to MEDIA-MTXT.

See record https://searchworks.stanford.edu/view/12377736

### DIFF
--- a/lib/locations_map.rb
+++ b/lib/locations_map.rb
@@ -62,7 +62,13 @@ class LocationsMap
   def load_home_locations
     CSV.parse(File.read(File.join(__dir__, 'translation_maps', 'locations.tsv')),
               col_sep: "\t").each_with_object({}) do |(home_location, library_code, folio_code), hash|
-      library_code = { 'LANE' => 'LANE-MED' }.fetch(library_code, library_code)
+      library_code = {
+        'LANE' => 'LANE-MED',
+        # Handle mapping of MEDIA-CRES to MEDIA-CENTER, which was MEDIA-MTXT in Symphony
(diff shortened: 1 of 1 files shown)


lib/locations_map.rb
```diff
@@ -64,6 +64,11 @@ def load_home_locations
               col_sep: "\t").each_with_object({}) do |(home_location, library_code, folio_code), hash|
       library_code = { 'LANE' => 'LANE-MED' }.fetch(library_code, library_code)
 
+      # Handle mapping of MEDIA-CRES to MEDIA-CENTER, which was MEDIA-MTXT in Symphony
+      # https://github.com/sul-dlss/folio_migration/blob/63166e3163bcee0562dd0c9b96d313e1adf968ac/mapping_files/temp_locations.tsv#L19
+      # https://stanfordlib.slack.com/archives/G018TJ20XGE/p1695672377659539
+      library_code = 'MEDIA-MTXT' if library_code == 'MEDIA-CENTER'
```
jcoyne, line 70: Is it possible to integrate this into the literal hash on line 65?

jcoyne (changes requested):
See comment about using the literal hash.

lib/locations_map.rb
thatbudakguy: updated

jcoyne (approved):

Issue #1237, opened 2023-09-12
Fix MHLD sorting for selecting the latest


cbeer:
Problematic examples:
https://searchworks.stanford.edu/view/364046
https://searchworks.stanford.edu/view/10694200 

VUF-9321
SW-4180

cbeer:
`Date.parse` is a little to... eager.. to extract some date:

```https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/folio/holdings.rb#L36

thatbudakguy:
Could be helpful to have a real database dump of enum/cron values so that we know the full scope of how bad/weird the data can be.

hudajkhan:
Closed by #1265

PR #1265, opened 2023-10-04
adding handling for dates separated by slashes

hudajkhan:
When faced with strings with slashes, (e.g. "Oct 25/31 2018" for https://searchworks.stanford.edu/view/10694200), the current Folio Holdings code that extracts the latest date based on Folio chronology/enumeration data uses Date.parse which can lead to incorrect results. 

What this pull request does:

- When the code sees a "/" in the date string, it splits the string and treats that as two separate values
- If there are multiple slashes in a string, we treat them in the following way: Oct/Nov 13/12 2022/2023 will translate to Oct 13 2022 and Nov 12 2023.  
- Because the objective is to get the latest or maximum date for sorting purposes, the code then retrieves these multiple strings, parses them or extracts a four digit year, and then returns the string corresponding to the maximum date.
- This pull request also takes the date parsing code and puts that code in its own class. The class name has been added to boot.rb to help with being able to refer to this class from the Holdings class.  (If there is another way to accomplish this reference, please let me know.)
- Furthermore, the original code would first use Date.parse, and then if that did not return a result or threw an error, would use a regex to scan for four digit years in the string. The parsing mechanism and the scanning mechanism have been separated out into their own methods. 
- The code also adds a check to the result of Date.parse to ensure the years returned are still 4 digit years starting with 18, 19, or 20.  This check is included to prevent situations where the Date.parse does not throw an error but returns with years like "20224" because the original string had a typo.
- Tests have been added to address the separate combinations of slashes in date strings this pull request is supposed to handle. 

What this pull request does not do:
- Whereas the chronology values seem date-centric, enumeration can have multiple values which are not related to dates (e.g. no. 297 or v. 13).  Enumeration values may still have years. Date.parse will not always throw an error and it can provide incorrect year results for some of these strings.  This code does not include any extra handling of those situations. 
- Sometimes date values have dashes e.g "Oct-Nov".  This code does not include any extra processing to extract "Oct" and "Nov" separately. 
- The logic for extracting tokens has been left as it was.

Addresses #1237


jcoyne (approved):

hudajkhan:
Thank you!

PR #1289, opened 2023-12-06
Send an explicit 0 count when we attempt a folio extraction but no records are returned

cbeer:


jcoyne:
Can you explain a bit more why we need/want to do this?

PR #1306, opened 2024-01-19
Allow non-geo content types to point to source collection

### LINKED ISSUE #1305: Geo indexer does not generate dc_source_s for scanned maps
Earthworks displays collection context widgets on the item detail page for both collections and the items that are members of a collection.

On a collection:
Screen Shot 2022-12-13 at 11 52 16

On a record from the collection:
Screen Shot 2022-12-13 at 11 58 33

Some collections exhibit this behavior correctly, but others do not. We need to figure out why this is, and if necessary remediate data or change behavior so that all collections display the widget consistently. @kimdurante's hypothesis is that it has to do with bounding boxes/geometry being missing on the collection or its members.

Example of a collection record where the widgets display **correctly**:
[China provinces data](https://earthworks.stanford.edu/catalog/stanford-nz187ct8959)

Example of a collection record where the widgets **do not display**:
[Caroline Batchelor maps](https://earthworks.stanford.edu/catalog/stanford-ct961sj2730)

thatbudakguy:
This fixes #1305, which will restore the relations widget in
Earthworks for scanned map content once it is reindexed.

### DIFF
--- a/spec/integration/geo_config_spec.rb
+++ b/spec/integration/geo_config_spec.rb
@@ -64,6 +64,10 @@ describe 'EarthWorks indexing' do
     it 'contains date' do
       expect(result['solr_year_i']).to eq [1603]
     end
+
+    it 'contains a reference to the source collection' do
+      expect(result['dc_source_sm']).to include('stanford-bf420qj4978')
+    end
   end
   context 'image map book content without dc:type' do
     let(:druid) { 'ny179kk3075' }
--- a/lib/traject/config/geo_config.rb
+++ b/lib/traject/config/geo_config.rb
@@ -402,7 +402,6 @@ to_field 'solr_year_i' do |record, accumulator|
 end
 
 to_field 'dc_source_sm' do |record, accumulator|
-  next unless record.dor_content_type == 'geo'
   next unless record.collections&.any?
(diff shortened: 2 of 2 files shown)


spec/integration/geo_config_spec.rb
```diff
@@ -64,6 +64,10 @@ def stub_mods_request(druid, body)
     it 'contains date' do
       expect(result['solr_year_i']).to eq [1603]
     end
+
+    it 'contains a reference to the source collection' do
```
jcoyne, line 68: Should there be a test for non-geo types?

jcoyne (changes requested):

spec/integration/geo_config_spec.rb
jcoyne: Cool. I was confused by this being geo_config_spec.

jcoyne (approved):

spec/integration/geo_config_spec.rb
thatbudakguy: That's fair – geo_config is more accurately "earthworks_config" since it handles a bunch of things that aren't strictly `geo`.

PR #1310, opened 2024-01-23
grab polineid from titles table inslucde of po_line table

dnoneill:


jcoyne:
Can you explain the motivation behind this change?

corylown:
@dnoneill could we set this PR to draft until we've done further testing/analysis?

cbeer:
I don't think `(titles.jsonb ->> 'instanceId')::uuid ` is an indexed column in the FOLIO database, but maybe/hopefully that changes in a new FOLIO release.

dnoneill:
@cbeer how do I check if ```(titles.jsonb ->> 'instanceId')::uuid ``` is an indexed field?

cbeer:
We huddled and confirmed that `(titles.jsonb ->> 'instanceId')` is still not indexed in postgres.

cbeer:
Confirming that FOLIO still doesn't have an index on `(titles.jsonb ->> 'instanceId') ` or `po_lines.jsonb ->> 'holdings_id'` (which is what the FOLIO UI uses... explaining why that section of the FOLIO page takes a long time.)

PR #1322, opened 2024-02-05
stop indexing collection if they aren't public

dnoneill:
From what I can tell this is working on sw. It will not fix the issue with currently indexed items but it will stop this issue from re-occuring. I have been using https://searchworks-preview-stage.stanford.edu/view/bh250xv2418 + https://argo-stage.stanford.edu/view/druid:qr592gj5093 to test that this works. It looks like you can release/withdraw a collection without its objects and the items will get reindex. Be warned, it can take a while to reindex.

Closes https://github.com/sul-dlss/SearchWorks/issues/3840

If a argo collection has not been released to Searchworks, the solr index with not provide the id in the collection field. It still provides collection_with_title which becomes -|- which I thought would be good to keep to let us know that there is a non public collection associated with the record but Searchworks doesn't display those non viewable collections.

### DIFF
--- a/lib/public_xml_record.rb
+++ b/lib/public_xml_record.rb
@@ -20,6 +20,8 @@ class PublicXmlRecord
   end
 
   def searchworks_id
+    return nil unless collection_in_searchworks?
+
     catkey.nil? ? druid : catkey
   end
 
@@ -31,7 +33,19 @@ class PublicXmlRecord
 
   # @return objectLabel value from the DOR identity_metadata, or nil if there is no barcode
   def label
+    return nil unless collection_in_searchworks?
+
     get_value(public_xml_doc.xpath('/publicObject/identityMetadata/objectLabel'))
+  end
+
+  def collection_in_searchworks?
+    return true unless collection?
+
+    if public_xml_doc.xpath('/publicObject/releaseData').present?
+      get_value(public_xml_doc.xpath('/publicObject/releaseData/release[@to="Searchworks"]')) == 'true'
+    else
+      false
+    end
   end
 
   def get_value(node)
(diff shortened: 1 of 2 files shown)


dbranchini:
@jcoyne I think that's more of a question for @andrewjbtw.

andrewjbtw:
>to not show any collection information at all

Are you referring to the collection information being in the MODS? I don't know how we can prevent that from appearing. My understanding of the scope of this issue is to prevent displaying links to collection pages that don't exist.

lib/public_xml_record.rb
```diff
@@ -31,9 +33,21 @@ def catkey
 
   # @return objectLabel value from the DOR identity_metadata, or nil if there is no barcode
   def label
+    return nil unless collection_in_searchworks?
+
     get_value(public_xml_doc.xpath('/publicObject/identityMetadata/objectLabel'))
   end
 
+  def collection_in_searchworks?
```
cbeer, line 41: I think I've been tripped up by this method name and/or the context for how this gets used. Maybe it's just the  method name, but I wonder if there's a more explicit/clearer way to implement this (kind of confusing in the first place) requirement.

From your description, it sounds like we're trying to suppress the id and title from the `collection` and `collection_with_title` fields?

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/sdr_config.rb#L295-L303

Instead of tweaking the underlying data for collection objects (which.. feels like it might be useful in some contexts.. debugging if nothing else?), what if `PublicXmlRecord` had a `released_collections` accessor that we used anywhere we needed it?

dnoneill: That works. The upside of this is we don't have to make any changes to the code in searchworks or the solr index. I have also left collection_with_title to still index with this it just returns an empty value of -|- which I thought would be good for debugging. Basically it would indicate that this has a collection attached but it isn't public. Probably not the best indicator so I can update the code, but I think that means we would need to update the solr index along with Searchworks.

dnoneill:
closing. This is too old to work correctly.

PR #1331, opened 2024-02-07
Add topic values to bib_search

cbeer:
At least some of the sw_index_tests expect to search for words from the title + topics 🤷‍♂️


jcoyne (changes requested):
Can we have a test for these changes?

cbeer:
It strikes me as premature to add tests for the new `bib_search` fields. We're still iterating on what's necessary in this field after setting `sow=false`, and there's probably several more round of iteration before we get the right data here.

PR #1341, opened 2024-02-20
[refactor] Make it easier to see where the call_number comes from

jcoyne:


cbeer:
See #1352 as an alternative. I don't think we need to introduce a new property, because we already have the `type` value.... and we don't have any need for call numbers for these placeholder e-resource items anyway.

jcoyne:
@cbeer I merged yours and put the remainder of this one on top. I think this still addresses the main point I had, that `call_number` should not be a parameter.

PR #1345, opened 2024-02-21
Don't index summary_struct when there is no data

jcoyne:
This avoids useless data like this:

```ruby
[{:fields=>[], :label=>"Summary", :unmatched_vernacular=>[]}, {:fields=>[], :label=>"Content advice", :unmatched_vernacular=>[]}]
```

As seen on records like https://searchworks.stanford.edu/view/2472159

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -1582,10 +1582,8 @@ to_field 'toc_struct' do |marc, accumulator|
   end
 
   unmatched_vern_fields = get_unmatched_vernacular(marc, '505')
-  unless unmatched_vern_fields.nil?
-    unmatched_vern_fields.each do |vern_field|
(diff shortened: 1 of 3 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -1649,7 +1648,7 @@ def accumulate_summary_struct_fields(matching_fields, tag, label, marc, accumula
   end
 
   accumulator << { label:, fields:,
-                   unmatched_vernacular: unmatched_vern } unless fields.empty? && unmatched_vern.nil?
+                   unmatched_vernacular: unmatched_vern } unless fields.empty? && unmatched_vern.empty?
```
cbeer, line 1651: The `unless` is pretty confusing. Is this the same thing?

```suggestion
                   unmatched_vernacular: unmatched_vern } if fields.empty? && !unmatched_vern.empty?
```

marlo-longley (approved):
This worked to fix the given example of "World: 1" and kept the `summary_struct` for records with good data that I tested.

PR #1354, opened 2024-02-23
Use 'items' to refer to the physical things on shelves

cbeer:
This aligns us a little better with FOLIO's use of holdings, although our notion of items isn't quite FOLIO's (but it's way more like a FOLIO item than a FOLIO holdings record).


corylown (approved):
🎉

PR #1355, opened 2024-02-23
Use the stdlib rather than DIY

jcoyne:


cbeer:
I think there is more wrong with that fallback beyond `filter`/`reject`. I reviewed the solrmarc code and[ associated documentation](https://consul.stanford.edu/display/NGDE/Items+Needing+Special+Discovery+Treatment), and I think this PR will get us closer to the intent: https://github.com/sul-dlss/searchworks_traject_indexer/pull/1357

PR #1364, opened 2024-02-27
Add effective_permanent_location_code to the FOLIO item json

cbeer:


jcoyne (approved):

lib/folio_item.rb
```diff
@@ -138,6 +138,7 @@ def to_item_display_hash
       instance_id: instance&.dig('id'),
       instance_hrid: instance&.dig('hrid'),
       # FOLIO item data to replace library/home_location/current_location some day
+      effective_permanent_location_code: display_location&.dig('code'),
```
jcoyne, line 141: Can we use `display_location_code` here instead?

cbeer, line 141:
```suggestion
      effective_permanent_location_code: display_location_code,
```

jcoyne (approved):

PR #1367, opened 2024-02-27
Support index maps with .geojson extension

thatbudakguy:
See https://github.com/sul-dlss/gis-robot-suite/issues/817


lwrubel (approved):

thatbudakguy:
tested in stage with https://earthworks-stage.stanford.edu/catalog/stanford-wd901gk5706 and confirmed that the `.geojson` extension works.

PR #1378, opened 2024-03-01
Filter course reserves to courses that end in the future

### LINKED ISSUE #1376: Old courses are being indexed
For example https://searchworks.stanford.edu/catalog?f%5Bcourses_folio_id_ssim%5D%5B%5D=142fe167-95fb-42c6-8660-152eb66ae2af displays course that ended `2023-12-18`

jcoyne:
Otherwise Searchworks shows items on course reserves forever. Fixes #1376

### DIFF
--- a/lib/traject/readers/folio_postgres_reader.rb
+++ b/lib/traject/readers/folio_postgres_reader.rb
@@ -168,6 +168,7 @@ module Traject
       end
     end
 
+    # Retrieve the courses that end sometime in the future
     def course_reserves
       @course_reserves ||= begin
         response = @connection.exec <<-SQL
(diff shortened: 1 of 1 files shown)


corylown:
@jcoyne the [example given](https://searchworks.stanford.edu/catalog?f%5Bcourses_folio_id_ssim%5D%5B%5D=142fe167-95fb-42c6-8660-152eb66ae2af) in #1376 no longer displays any courses. Did this get fixed elsewhere?

jcoyne:
@corylown yes, there's definitely been some data change, however I still think we need this filter so that we can prune the data in Searchworks without accidentally creating dangling pointers.

saseestone:
Confirming that we should filter Courses based on end date.  The end date should be in the future.  We don't have any need to display the course beyond the FOLIO end date.

PR #1379, opened 2024-03-01
Add callNumber, volume, enumeration, chronology to boundWith 

jcoyne:
This is required in order to implement the display changes called for in https://github.com/sul-dlss/SearchWorks/issues/3200

### DIFF
--- a/lib/traject/readers/folio_postgres_reader.rb
+++ b/lib/traject/readers/folio_postgres_reader.rb
@@ -373,7 +373,11 @@ module Traject
                                 'id', parentItem.id,
                                 'hrid', parentItem.jsonb ->> 'hrid',
                                 'barcode', parentItem.jsonb ->> 'barcode',
-                                'status', parentItem.jsonb #>> '{status, name}'
+                                'callNumber', parentItem.jsonb #>> '{effectiveCallNumberComponents, callNumber}',
+                                'status', parentItem.jsonb #>> '{status, name}',
+                                'enumeration', parentItem.jsonb->>'enumeration',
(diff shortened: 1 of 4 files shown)


lib/traject/readers/folio_postgres_reader.rb
```diff
@@ -362,7 +362,8 @@ def contents_sql_query(conditions, addl_from: nil)
                                 'hrid', parentInstance.jsonb ->> 'hrid',
                                 'title', parentInstance.jsonb ->> 'title'
                               ),
-                              'holding', jsonb_build_object('effectiveLocationId', parentHolding.jsonb ->> 'effectiveLocationId'),
+                              'holding', jsonb_build_object('effectiveLocationId', parentHolding.jsonb ->> 'effectiveLocationId',
+                                                            'callNumber', parentHolding.jsonb ->> 'callNumber'),
```
cbeer, line 366: What part of https://github.com/sul-dlss/SearchWorks/issues/3200 needs the holdings call number? I think we actually need the item call number (plus the volume, enumeration, etc) to get the call number Darcy's showing in:

Screenshot 2024-03-04 at 07 51 16

At least, in the example Darcy provided, in the librarian view the holding call number for https://searchworks.stanford.edu/view/2307563 is just the base call  number:
Screenshot 2024-03-04 at 07 52 12

jcoyne: @cbeer Thank you. I also wonder if this would be better if we added a `boundWith` into the `item_display_struct`.

corylown (approved):

Issue #1343, opened 2024-02-20
(Determine if we can) map FOLIO instance data to the MARC leader


cbeer:
In Symphony, our MARC leaders were always populated with the format information (e.g. "This thing is a book"). In FOLIO, we're not mapping anything from the FOLIO instance metadata into the MARC leader, leading to some funny behavior where we expected there'd be required data (e.g. https://github.com/sul-dlss/SearchWorks/issues/3970).

Is there data we can + should map into the MARC leader in [MarcRecordInstanceMapper](https://github.com/sul-dlss/searchworks_traject_indexer/blob/79053bcebd6089737b2e232c7689919d22bb1605/lib/folio/marc_record_instance_mapper.rb#L9)? https://folio-org.atlassian.net/wiki/spaces/FOLIOtips/pages/5673766/Default+MARC+Bibliographic-to-Inventory+Instance+Mappings indicates that FOLIO maps the MARC leader byte 7 information to FOLIO instance metadata.

1) Inventory what MARC leader data we're using in the indexer
2) Talk with Alissa et al about what/where/how to map data from the instance to FOLIO
3) Implement the mapping when we're constructing our own MARC records from FOLIO instance metadata.

hudajkhan:
May also relate to #3970 (SearchWorks)

taylor-steve:
In the indexer we are:

Building `format_main_ssim` from:
- [leader/06, some cases considering 008/26 and 008/33](https://github.com/sul-dlss/searchworks_traject_indexer/blob/7e39606213835bb2f8c546a8f569a145aa5dccc0/lib/traject/config/folio_config.rb#L1018)
- [leader/07 == 's' with 008/21 present](https://github.com/sul-dlss/searchworks_traject_indexer/blob/7e39606213835bb2f8c546a8f569a145aa5dccc0/lib/traject/config/folio_config.rb#L1068)
- [leader/07 == 'i' with 008/21 present](https://github.com/sul-dlss/searchworks_traject_indexer/blob/7e39606213835bb2f8c546a8f569a145aa5dccc0/lib/traject/config/folio_config.rb#L1101)
[...]

PR #1387, opened 2024-03-04
Map FOLIO data to the MARC leader

taylor-steve:
Addresses #1343.

Draft until Jeanette can review the mapping as requested.

Calling out we *could* drop the query modifications here and do the comparisons via instance type/mode of issuance UUID directly (the UUIDs for both are in the existing instance data) if we wanted to prioritize query simplicity.


cbeer:
>  do the comparisons via instance type/mode of issuance UUID directly 

I like using the names to match; we ran into some UUID instability across environments during the initial FOLIO work.

PR #1389, opened 2024-03-04
Index bound with parent item ids

jcoyne:
For drawing the modal in https://github.com/sul-dlss/SearchWorks/issues/3200

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -2571,6 +2571,11 @@ to_field 'holdings_json_struct' do |record, accumulator|
   }
 end
 
+# This allows Searchworks to query the boundWith children.
+to_field 'bound_with_parent_item_ids_ssim' do |record, accumulator|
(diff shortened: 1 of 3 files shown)


lib/traject/config/folio_config.rb
```diff
@@ -2600,6 +2600,10 @@ def accumulate_summary_struct_fields(matching_fields, tag, label, marc, accumula
   }
 end
 
+to_field 'bound_with_parent_item_ids_ssim' do |record, accumulator|
+  accumulator << record.index_items.filter_map { |holding| holding.bound_with_holding&.dig('boundWith', 'item', 'id') }
```
cbeer, line 2604:
```suggestion
  accumulator.concat record.index_items.filter_map { |holding| holding.bound_with_holding&.dig('boundWith', 'item', 'id') }
```

Otherwise we'll get `[['A', 'B', 'C']]`, right?

Should this have a test?

Issue #1380, opened 2024-03-01
Create a traject reader that provides cocina data from purl-fetcher


thatbudakguy:
Prerequisite for #770, and for more general future work to index using cocina instead of public XML.

- [x] Check what existing libraries offer logic for accessing Cocina fields (`cocina-models`?)
- [x] Create a class that bundles cocina-related accessors, parallel to `PublicXmlRecord` but for JSON
- [x] Create new versions of the `KafkaPurlFetcherReader` and `DruidReader` that yield the new `CocinaRecord` class

PR #1391, opened 2024-03-05
Add dor-services-client in order as a prerequisite for accessing cocina

aaron-collier:
This is a prerequisite for #1380 in order to find cocina objects for metadata indexing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,4 +42,5 @@ gem 'activesupport', '~> 7.0'
 gem 'slop'
 
 gem 'dor-event-client'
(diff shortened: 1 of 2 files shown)


jcoyne:
I'm not sure I understand why we need to add this library?  Perhaps it would be helpful if you added some new functionality that used this library so we could see what you're trying to accomplish.

aaron-collier:
Perhaps this should be a chat that includes @thatbudakguy as we were under the impression we needed access to cocina objects for this indexing.

PR #1398, opened 2024-03-06
Add initial basic PublicCocinaRecord class

aaron-collier:
As a prerequisite for serializing ardvark data (in order to upgrade earthworks) this creates the initial (and basic) PublicCocinaRecord class.

### DIFF
--- a/spec/lib/public_cocina_record_spec.rb
+++ b/spec/lib/public_cocina_record_spec.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe PublicCocinaRecord do
+  let(:druid) { 'bb021mm7809' }
+  let(:purl_url) { 'https://purl.stanford.edu' }
+  let(:public_cocina_record) { described_class.new(druid, purl_url:) }
+  let(:body) { File.new(file_fixture("#{druid}.json")) }
+
+  before do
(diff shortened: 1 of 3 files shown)


spec/lib/public_cocina_record_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'spec_helper'
+
+RSpec.describe PublicCocinaRecord do
+  let(:druid) { 'bb021mm7809' }
+  let(:purl_url) { 'https://purl.stanford.edu' }
+  let(:public_cocina_record) { described_class.new(druid, purl_url:) }
+
+  def stub_purl_request(druid, body)
+    stub_request(:get, "https://purl.stanford.edu/#{druid}.json").to_return(status: 200, body:)
```
jcoyne, line 11: How about just moving this into the before block and then we don't need to define a method?

spec/lib/public_cocina_record_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  before do
+    stub_purl_request(druid, File.read(file_fixture("#{druid}.json").to_s))
```
jcoyne, line 15: can you extract `let(:body) { File.new(file_fixture("#{druid}.json") }` (you don't need to read the file).

spec/lib/public_cocina_record_spec.rb
aaron-collier: Done.

jcoyne (approved):

PR #1409, opened 2024-03-08
[DRAFT | WIP] Expand public_cocina_record and add geo_aardvark_config

aaron-collier:
This is the next part of #1380 

This should be considered DRAFT until the current cocina->aardvark mapping is included. Refactoring and cleanup is expected before undrafting but comments are welcome to help guide the direction and patter (currently following: https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/geo_config.rb and https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/public_xml_record.rb)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,5 +41,6 @@ end
 gem 'activesupport', '~> 7.0'
 gem 'slop'
 
+gem 'cocina-models'
 gem 'dor-event-client'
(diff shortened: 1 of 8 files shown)


thatbudakguy:
closing in favor of #1448 which builds on this

PR #1418, opened 2024-03-12
Restore "replace half diacritics U+FE20, U+FE21 with full diactritic U+0361"

cbeer:
Restore https://github.com/sul-dlss/searchworks_traject_indexer/commit/a9947d45601c2615a70d30b1e037d97fc2758290

I think the original regex was over-matching, but the lookahead/behinds are too clever for me (it's probably more accurate, but this seems to pass the same tests🤷‍♂️)


cbeer:
What are you thinking? I already included a test for the Gale URL.

dnoneill:
My bad. I missed that.

dnoneill (approved):

PR #1439, opened 2024-05-08
Remove extraneous code

jcoyne:
A /thumb node will always exist in the XML if a suitable thumb is available. There is no need for this fallback, this was duplicating the behavior of the ThumbnailService in DSA


jcoyne:
@jcoyne this would not be needed after switching to cocina, since this is xml code.

PR #1448, opened 2024-05-31
Add Aardvark indexing config for Earthworks

### LINKED ISSUE #770: Generate Aardvark format records for geoblacklight
Blocked by #1380 

In versions after v4, geoblacklight uses the [Aardvark schema](https://opengeometadata.org/docs/ogm-aardvark) instead of the previous v1.0. In order to update Earthworks to v4 and beyond, we need to write records to solr in this format instead of the existing configuration in `geo_config.rb`.

This blocks https://github.com/sul-dlss/earthworks/issues/936.

See:
- [metadata mapping ISO19139 -> cocina -> Aardvark](https://docs.google.com/spreadsheets/d/1GV1QK04d870EM5ismR0iHef-9n5bIe5laP0VVjzq0FA/edit#gid=2042286233)
- [OGM notes about mapping between v1 format and Aardvark](https://opengeometadata.org/aardvark-gbl-1-crosswalk/)

After the mapping is settled, we need to:
- [x] move our existing 1.0 metadata in OGM to a v1.0-specific directory (or just delete it, since it won't get updated)*
- [ ] update [the script that populates our OGM repo](https://github.com/OpenGeoMetadata/edu.stanford.purl/blob/master/Rakefile) to publish Aardvark to an aardvark directory

*as of geoblacklight community meeting 1/10/24, there seems to be a consensus that there's no need to keep old/v1 records updated once you move to aardvark

### LINKED ISSUE #1337: Update EarthWorks records to point directly to shelved metadata in stacks
We populate `dct_references_s` for our geoblacklight records with several references to external metadata formats:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/ef2228b4ae87f22cc2ec8adda8624a5d6b923018/lib/traject/config/geo_config.rb#L330-L351

We had been using a strategy of unzipping and committing these to our OGM repo, and then referencing the GitHub versions. We're no longer doing that as of https://github.com/OpenGeoMetadata/edu.stanford.purl/pull/131, and as of https://github.com/sul-dlss/gis-robot-suite/issues/673 we can just reference the versions that will be in stacks.

Similarly, as of https://github.com/sul-dlss/gis-robot-suite/issues/753, the `data.zip` and other files won't exist. Instead, we should reference the whole-object download URL in stacks: `https://stacks.stanford.edu/object/[druid]`, which downloads a .zip of everything automatically.

### LINKED ISSUE #1380: Create a traject reader that provides cocina data from purl-fetcher

thatbudakguy:
Based on (& supersedes) #1409.

This adds some new methods to PublicCocinaRecord, but mostly writes a bunch of new Traject macros used to extract things from Cocina.

I did it this way because:
- a lot of common operations were repeated in the indexer config otherwise
- it made the indexer config a lot shorter and more readable
- it was easier to write unit tests for small, repeated bits of logic in addition to the config as a whole

The hope is that when we rewrite our "SDR indexing to SearchWorks" config, we can take advantage of some of these macros.

I deployed this to staging and tested releasing/unreleasing from argo-stage and it worked as expected; records made it to earthworks-stage (using the new Aardvark core). I have also used it to index all 30,000+ items released to Earthworks into http://earthworks-uat.stanford.edu, where you can see them currently.

Closes #1337 
Closes #1380 
Closes #770


lib/public_cocina_record.rb
```diff
@@ -3,27 +3,77 @@
...
 
-  def initialize(druid, purl_url: 'https://purl.stanford.edu')
+  def initialize(druid, public_cocina, purl_url: 'https://purl.stanford.edu')
     @druid = druid
     @purl_url = purl_url
+    @public_cocina_doc = JSON.parse(public_cocina)
+  end
+
+  def public_cocina
+    @public_cocina ||= self.class.fetch("#{purl_url}/#{druid}.json")
```
cbeer, line 20: What's the relationship between `@public_cocina_doc` and `@public_cocina`?

lib/public_cocina_record.rb
```diff
@@ -3,27 +3,77 @@
...
+  end
+
+  def cocina_access
+    @cocina_access ||= Cocina::Models::DROAccess.new(public_cocina_doc['access'])
```
jcoyne, line 24: The reason I didn't use the cocina structures here is that it means we need to keep in sync with the version of cocina models.   It's safer to just use the JSON in case the data doesn't match the structures.

lib/public_cocina_record.rb
thatbudakguy: Hmm. this might be a holdover from #1409. I'm not sure why you'd ever have the thing `fetch` itself. Seems like we only need `public_cocina_doc`?

thatbudakguy: Tests passed without this method, so I removed it.

lib/public_cocina_record.rb
thatbudakguy: Would we want to use that as a warning that stuff needs remediation, and avoid indexing it entirely? I guess I didn't realize that things could get substantially out of sync. That seems like the problem that cocina-models is supposed to solve.

thatbudakguy: Added a commit to remove all the cocina-models usage except for `TitleBuilder`, which seemed non-trivial, and one instance of constructing fixture data in a test.

jcoyne (changes requested):

PR #1484, opened 2024-07-03
Stop mapping home_location and current_location; fixes #1348

### LINKED ISSUE #1348: Clean up Symphony-ism `home_location` + `current_location`
With https://github.com/sul-dlss/SearchWorks/pull/3892/, we're no longer using Symphony codes from the  `current_location` field. We should be able to stop using those attributes, clean up any lingering symphony-isms, and consolidate on `temporary_location`, `display_location`, and `status`.

cbeer:


jcoyne (commented):
spec/lib/folio_holding_spec.rb:150 references `home_location` still

jcoyne (approved):

PR #1490, opened 2024-07-09
add searchworks url to earthworks solr doc

peetucket:
Part of https://github.com/sul-dlss/earthworks/issues/310

Show link to searchworks record in Earthworks if released to searchworks.  Examine public meta json when indexing to discover if released to searchworks and then construct URL for solr doc.

To do:
- [x] tests
- [x] test on server


lib/traject/macros/meta_json.rb
thatbudakguy, line 1: I don't think this needs its own file. I'd put it in `Traject::Macros::Cocina`.

thatbudakguy: (I realize the meta.json info is no longer technically in the cocina, but it seems related enough)

peetucket: yeah, ok -- not sure how pure we wanted to be, but technically release tags were never in cocina (they were injected into public XML before but are stored elsewhere)

peetucket: but having said that, the searchworks_url macro doesn't really care about where the release knowledge came from when it was added to the record object, so agreed it probably doesn't need it's own file

thatbudakguy (approved):

PR #1505, opened 2024-08-14
Geo Aardvark: skip records with no bounding box

### LINKED ISSUE #1504: Do not index geo material without a bounding box
In the old config, we skipped records that did not have valid coordinates in the `solr_geom` field:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/4ed5e9e8fee1100b26d4c48262e91dcbfa66fc47/lib/traject/config/geo_config.rb#L465-L475

In the new config, there are two fields, `locn_geometry` and `dcat_bbox`. Because we always use the same `ENVELOPE` syntax for both, we populate `locn_geometry` and then set `dcat_bbox` to the same value as `locn_geometry`, since they will never differ (for Stanford records, anyway).

Currently, we allow both fields to be empty, because [the spec](https://opengeometadata.org/ogm-aardvark/#bounding-box) for Aardvark has both as "suggested" fields rather than required. This is a departure from the v1 schema, which required a bounding box.

As per discussion at the GIS team meeting 8/12, we don't think it's useful to support having content with no geospatial information at all (bounding box or geometry) in Earthworks, even if the spec allows it. Thus, we want to bring back the behavior from the old config of skipping these records and publishing a warning message to SDR to indicate that they were skipped because of a lack of valid coordinates.

To do this, we can insert an `each_record` block that skips records with no `locn_geometry` (no need to check `dcat_bbox` since they will always be identical). The block can either come immediately after the field definition or at the end. It'll probably look pretty similar to the block from the old config linked above.

marlo-longley:
Closes #1504

### DIFF
--- a/lib/traject/config/geo_aardvark_config.rb
+++ b/lib/traject/config/geo_aardvark_config.rb
@@ -314,6 +314,14 @@ unless settings['writer_class_name'] == 'Traject::DebugWriter'
   end
 end
 
+each_record do |record, context|
+  next if context.output_hash.key?('locn_geometry')
+
+  message = 'No geometry available for item'
+  SdrEvents.report_indexing_skipped(record.druid, target: settings['purl_fetcher.target'], message:)
+  context.skip!("#{message}: #{context.output_hash['id']}")
+end
+
 # Log the indexing time for each record
 each_record do |record, context|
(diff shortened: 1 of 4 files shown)


thatbudakguy (approved):
perfect!

PR #1509, opened 2024-08-26
Revert "Geo Aardvark: Reverse the boolean logic for gbl_georeferenced_b"

thatbudakguy:
This reverts commit 91e574a9e967d2b8af87252b51bedb86645b361b.

### DIFF
--- a/spec/integration/geo_aardvark_config_spec.rb
+++ b/spec/integration/geo_aardvark_config_spec.rb
@@ -271,7 +271,7 @@ RSpec.describe 'EarthWorks Aardvark indexing' do
     end
   end
 
-  context 'with the georeferenced map version of a scanned map' do
+  context 'with a scanned map that was georeferenced' do
     let(:druid) { 'kd514jp1398' }
 
     it 'maps the resource class' do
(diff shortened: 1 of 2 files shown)


spec/integration/geo_aardvark_config_spec.rb
```diff
@@ -296,18 +296,6 @@
...
   end
 
-  context 'with a scanned map that has a georeferenced map' do
-    let(:druid) { 'gm036df2527' }
-
-    it 'maps the georeferenced status' do
-      expect(result['gbl_georeferenced_b']).to eq true
-    end
-
-    it 'populates the link to the georeferenced map in dct_relation_sm' do
```
marlo-longley, line 306: Is dct_relations_sm modified in this PR? Might be missing something.

thatbudakguy: No, but this change was apparently part of the commit that got reverted, so it also was reverted...I'll see if I can fix that with another commit

marlo-longley: Well then it's my bad for a messy commit! We can add it later if easier with the git.

thatbudakguy: I was able to fix it and restore the fixture for this test also. Makes the PR a little cleaner to review.

marlo-longley (approved):

PR #1511, opened 2024-08-27
Add a reader for harvesting directly from purl-fetcher HTTP API

thatbudakguy:
This adds a traject reader that can be useful in development when
you want to quickly index many records from purl-fetcher without
resorting to Kafka. It is intended for dev use only.

It can point at any release target (searchworks, earthworks) and
index all of the items currently released to that target.

This PR also modifies PublicCocinaRecord and PublicXmlRecord to
optionally accept a connection object, so that a single Faraday
connection can be shared amongst the reader and records, which
enables parallelizing record-fetching from purl to match the
number of traject threads.

This setup allows indexing everything released to Earthworks in
a little under 5 minutes with 4 threads on my machine.


jcoyne:
@thatbudakguy Can you explain how would you use the purl-fetcher in development? This seems like functionality that is already built into Argo, and we're building it a second time, making more code paths to have to maintain.

thatbudakguy:
I use this to do reindexing of SDR content from my local machine. The past few times I've needed to do a full reindex of Earthworks, in both staging and prod, I've used it.

I also heard from some folks on the Earthworks workcycle who were working on solr changes that they wanted to index a lot of records (thousands) locally for testing. I think this is probably the easiest way to do that, or to create a local copy of all of production in order to test indexing changes.

For all of the cases I've used this, the underlying data in Argo has not changed – only the indexing configuration. Republishing the data there would move a lot of objects through the system unnecessarily.

jcoyne:
@thatbudakguy When you do a "Republish" in Argo, all it's basically just ensuring the metadata is fresh and telling the indexer to go.  It's not a lot of "unnecessary moving".  But if this is what you need then go for it.

PR #1513, opened 2024-08-28
Guard against missing dates in cocina JSON

### LINKED ISSUE #1512: Guard against empty cocina timestamps
Some of the logic in the geo indexing config uses a series of fallbacks to generate a "modified date" for the metadata itself:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/bbb0e6c0d98d829824946925335e3f2a6ccb797c/lib/traject/config/geo_aardvark_config.rb#L260-L265

Releasing the object `cb601kb6593` from argo-stage will throw the following error in the event log:
```json
{
  "host": "sw-indexing-stage-a.stanford.edu",
  "target": "Earthworks",
  "context": {
    "record": "<record #1 (/dev/null #1), output_id:stanford-cb601kb6593>",
    "index_step": "(to_field \"gbl_mdModified_dt\" at ./lib/traject/config/geo_aardvark_config.rb:264)"
  },
  "message": "no implicit conversion of nil into String",
  "invoked_by": "indexer"
}
```

thatbudakguy:
Closes #1512


jcoyne (changes requested):
The records you have are from 2 different schemas. One is a Cocina::DRO, and the other is a Cocina::DROWithMetadata.  This indicates a bigger problem with which schema should be expected.

thatbudakguy:
It's https://argo-stage.stanford.edu/view/druid:cb601kb6593.

jcoyne:
@thatbudakguy I republished the data to purl and it should have those fields now.

thatbudakguy:
Just encountered this again with https://sul-purl-stage.stanford.edu/yx324nv8969. Had to add fake created/modified dates to it as part of #1516 to avoid errors.

jcoyne:
@thatbudakguy it's just due to some bad data on stage.  When you encounter it, go to argo for the object and republish it.

Issue #1514, opened 2024-08-29
Geo Aardvark: schema changes


marlo-longley:
- [x] Only georeferenced maps should have the class 'Maps.' Currently, all raster data are classified as Maps (ex. Bathymetric GeoTIFFs such as this one: https://earthworks-stage.stanford.edu/catalog/stanford-wc605vp7672)
       The mapping logic for gbl_resourceClass_sm should be the same as it is for the gbl_georeferenced_b element:
               IF title.value CONTAINS '(Raster Image)' > Resource Class = Maps

- [x]  For dc_source_sm element: Only create this element if the item has a PURL. Apologies for not catching this sooner, but there are some georeferenced maps that have been sourced from other applications (such as the David Rumsey Map Collection site) and the identifiers that are generated will be invalid, see the dc_source_sm element this example:https://earthworks-stage.stanford.edu/catalog/stanford-yx324nv8969/raw

- [x] Update dcat_theme_sm mapping (see below)

kimdurante:
Hi Again. I cannot edit this ticket, so I'm adding this here in the comments. 
Because the Aardvark schema uses modified terms for some ISO topic category values, I would recommend the following for mapping for the `dcat_theme_sm` element:

IF subject*.source.code = ISO19115TopicCategory AND subject*.value = Climatology, Meteorology and Atmosphere > Climate

IF subject*.source.code = ISO19115TopicCategory AND subject*.value = Planning and Cadastral > Property

IF subject.type=topic AND subject.value = Aerial photographs OR Aerial surveys OR Bathymetric maps OR California as an island--Maps OR Composite map OR Landsat satellites OR Multibeam mapping OR Nautical charts OR Remote sensing OR Remote sensing images OR Satellite image maps > Imagery

IF subject.type = topic AND subject.value = Coasts OR Forests OR Forest canopies OR Forests and forestry OR Hydrology OR Hydrography OR Land cover OR Landforms OR Vegetation mapping > Land cover

OTHERWISE IF subject*.source.code=ISO19115TopicCategory > subject*.value


I hope this is not too confusing, happy to chat about it. This one is not a huge priority for me due to the fact that subjects and themes are inconsistently applied, and also that metadata for scanned maps do not include ISO topics, so a complete mapping to the dcat_theme element will require much more analysis, especially as we add more maps to EarthWorks

PR #1515, opened 2024-08-29
Ensure only georeferenced rasters are both "Maps" and "Datasets"

thatbudakguy:
Part of #1514

This changes the indexing strategy for resource class to use a
MatchMap to make it much simpler to get map/dataset values out
of the data.

It also ensures that any georeferenced images are counted as both
maps and datasets.


lib/translation_maps/geo_resource_class.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
marlo-longley, line 1: Wowww thanks MatchMap!

spec/integration/geo_aardvark_config_spec.rb
```diff
@@ -320,7 +319,7 @@
     let(:druid) { 'bb021mm7809' }
 
     it 'maps the resource class' do
-      expect(result['gbl_resourceClass_sm']).to eq %w[Datasets Maps]
+      expect(result['gbl_resourceClass_sm']).to eq ['Datasets']
```
marlo-longley, line 322: Aren't we supposed to test that these are classed as both Maps and Datasets?

lib/translation_maps/geo_resource_class.rb
thatbudakguy: It's perfect for this! Bill clearly has done a lot of this type of thing, including in traject.

spec/integration/geo_aardvark_config_spec.rb
thatbudakguy: Nope – this one is not georeferenced, so it's only a dataset.

marlo-longley (approved):

PR #1520, opened 2024-09-04
fix searchworks url formation

### LINKED ISSUE #1519: Searchworks link for geo items doesn't use catkey when it is present
Our current logic for generating the searchworks link always uses the object's druid to construct a URL, if it has been released to searchworks:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/1970b005886803ba2625fa95f1c2ff1ea51f4966/lib/traject/macros/cocina.rb#L88-L93

But for some objects like https://purl.stanford.edu/hj948rn6493 with a catkey, the correct searchworks URL uses the catkey instead of the druid: https://searchworks.stanford.edu/view/10784397.

That results in the "view in searchworks" link in earthworks going to a 404.

dnoneill:
closes #1519

### DIFF
--- a/lib/traject/macros/cocina.rb
+++ b/lib/traject/macros/cocina.rb
@@ -88,7 +88,7 @@ module Traject
       # Generate a Searchworks URL if object is released to Searchworks
       def searchworks_url
         lambda do |record, accumulator, _context|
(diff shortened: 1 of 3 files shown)


jcoyne:
Please add a test for this change.

dnoneill:
@jcoyne Done.

thatbudakguy (approved):

PR #1553, opened 2025-03-06
Stop indexing unneeded stanford_rights_metadata_s field

jcoyne:
It is no longer needed since https://github.com/sul-dlss/earthworks/commit/8c9226a53ba5fcee1907162d9cd0f6d89006b300


thatbudakguy (approved):
Fwiw our current config is https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/geo_aardvark_config.rb – to my knowledge this config is not used at all and can be deleted now.

PR #1565, opened 2025-03-20
Add 'Loose-leaf' as a format

cbeer:
- [ ] [add icon support in Searchworks](https://github.com/sul-dlss/SearchWorks/pull/4761)

### DIFF
--- a/spec/lib/traject/config/format_main_spec.rb
+++ b/spec/lib/traject/config/format_main_spec.rb
@@ -1775,7 +1775,7 @@ RSpec.describe 'Format main config' do
       end
     end
 
-    it 'is a Book' do
+    it 'is a Loose-leaf' do
       expect(result[field]).to eq ['Loose-leaf']
     end
   end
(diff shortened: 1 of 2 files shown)


spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1776,11 +1776,11 @@
     end
 
     it 'is a Book' do
```
jcoyne, line 1778:
```suggestion
    it 'is a Loose-leaf' do
```

spec/lib/traject/config/format_main_spec.rb
```diff
@@ -1789,7 +1789,7 @@
...
```
jcoyne, line 1791:
```suggestion
    it 'is a Loose-leaf' do
```

dbranchini:
I see two possibilities in the streamline icon set:

Screenshot 2025-04-10 at 12 47 07 PM
Screenshot 2025-04-10 at 12 45 46 PM

cbeer:
👍 note-paper makes sense to me

dnoneill (approved):

PR #1581, opened 2025-04-04
856$z is a repeating field

cbeer:


lib/marc_links.rb
```diff
@@ -37,7 +37,7 @@ def as_h
         link_title:,
         additional_text:,
         material_type: field['3'],
-        note: field['z'],
+        note: subzs,
```
dnoneill, line 40: I don't have a problem with this but I am wondering why the finding_aid? field isn't using this?

jcoyne, line 40: Would we want this joined, or is it better to leave it as structured data (and rename to `notes:`)?

cbeer: I'm not sure (and one of the reasons I merged #1579 as is). It seems pretty funny that we'd only look at the first `$z` 🤷‍♂️

cbeer: I'm not sure; I don't think we have any other use cases for this data yet, and the concatenated field is type-compatible with the existing data 🤷‍♂️

PR #1585, opened 2025-04-07
add meta_json data to collection

dnoneill:


jcoyne:
Can you say what you are fixing about the schema_dot_org_struct in the commit message?

lib/traject/config/folio_config.rb
```diff
@@ -2289,14 +2289,15 @@ def accumulate_summary_struct_fields(matching_fields, tag, label, marc, accumula
...
+    meta_json = PublicMetaJsonRecord.fetch(druid).public_meta_json_doc
     accumulator << {
       source:,
       item_type:,
       type: 'collection',
       druid:,
       id:,
       title:
-    }
+    }.merge(meta_json)
```
jcoyne, line 2300: This causes two extra requests to PURL per object indexed.  Do we need this in the index?  Can't we retrieve it for the collection itself (from the index) at view time?

dnoneill: You are suggesting in searchworks, in order to determine if we should display the earthworks links, we should query the meta_json?

jcoyne: Yes

jcoyne: @cbeer made the point in slack that the folio_config shouldn't need to hit purl (which the PublicMetaJsonRecord does).

Issue #1562, opened 2025-03-14
Add shelfkeys for Gov Docs/Sudoc call numbers for browse nearby


taylor-steve:
Investigation needed? Maybe more info coming from Gov Doc staff later.

https://www.fdlp.gov/cataloging-and-classification/classification-guidelines

--

The basic gist (hypothesized, for govdocs + caldocs):

- letter case doesn't matter.
- sometimes there's information in brackets that precede the callnumber (e.g. caldoc has `[folio] `). Ignore those strings and leading whitespace.
- there's one or more letters to start the call number. These are sorted alphabetically.
- there may be a space or a dot or nothing at all. It doesn't matter for sorting. 
- there's a number, sorted numerically (1, 2, 10, 52, 253). There may be a dot separating two numbers (e.g. sudocs). If so, each are sorted numerically (i.e. it's a separator, not a decimal)

## Caldocs

Caldocs have a Cutter-like identifier next (possibly a space, probably a dot, a letter, possibly a space, and a number). The letters are sorted alphabetically, the number numerically. There may be additional letters appended (in which case, nothing sorts before the letters). 

There's additional text after that. Continue sorting letters alphabetically and numbers numerically, with some exceptions:

- ignore whitespace, dots, and other symbols (except as a separator for numbers) 
- it might be a Roman numeral (or a Roman numeral followed by some other text). Ideally, IX sorts after VIII and before X
- it might be a 2 digit year (e.g. 00 should sort after 99), probably followed by a dash.

## SuDocs 

Sometimes the base call number then has a slash followed by some numbers or letters. The normal stem (without slashes) sorts before slashed data.

There's usually(?) a colon after the base or slashed call number.

Following the colon, there might be a year, some letters, or a number. If it's a year in the 19xx (or possibly 18xx?), it may be 3 digits because the leading 1 is implied. After 2000, it should be a 4 digit year. Use the full year for sorting.

When sorting whatever that is, data with years come first, followed by letters, followed by numbers. Supposedly, there might also be "words" that sort last of all.

After that, it's somewhat unclear what should happen. Probably again sort alphabetically but treat all numbers numerically? There might be 3 digit years (preceded by slashes) too.

## UN docs

There are groups of strings separated by slashes. Within those strings might be numbers. Numbers should be sorted numerically.  There might be some Roman numerals 🤷‍♂

taylor-steve:
[...]

PR #1586, opened 2025-04-08
Generate Sudoc shelfkeys

taylor-steve:
Part of #1562 

Adds support for SuDoc shelfkeys for browse nearby.

For the shelfkey I'm ordering the entire call number and there is quite the diverse set. Ultimately it isn't doing anything that fancy beyond trying to determine what is likely a year versus just a number. "Words" sort alphabetically. That seems to be the norm and they tend to be very late in the call number, in which case you rarely see them because they are consolidated behind the lopped call number. I think user feedback would be extremely valuable. I suspect there could be a range of practices that users would discover. It's a challenge to sample 2+ million call numbers.

As opposed to shelfkeys, here lopped call numbers are depending on the (mostly) consistent stem + book number part of the suffix. This seemed correct, in practice, but we might need to depend on user feedback for this as well.

### DIFF
--- a/lib/call_numbers/other_shelfkey.rb
+++ b/lib/call_numbers/other_shelfkey.rb
@@ -15,8 +15,6 @@ module CallNumbers
 
     # this transfomation only applies when generating shelfkeys
     def shelfkey_scheme
-      return 'sudoc' if scheme == 'SUDOC'
-
       'other'
     end
   end
--- a/lib/call_numbers/sudoc_shelfkey.rb
+++ b/lib/call_numbers/sudoc_shelfkey.rb
@@ -1,0 +1,144 @@
+# frozen_string_literal: true
+
+module CallNumbers
+  class SudocShelfkey < ShelfkeyBase
+    DELIMITER_PRECEDENCE = {
+      ':' => 'a',
+      '.' => 'b',
+      ';' => 'c',
+      '/' => 'd',
+      '+' => 'e'
+    }.freeze
+
+    TYPE_PRECEDENCE = {
+      alphabetic: 'q',
+      year: 'r',
+      numeric: 's'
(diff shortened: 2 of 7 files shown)


taylor-steve:
Ah that makes sense, it was literally configured the other way. I was referencing too many tutorials that weren't the ticket...

I do see some libraries recommend sorting "words" last. We could do some version of this I think, but it doesn't seem super impactful to me as they most frequently appear at the end of the call number. What do you think? I could check back with the librarians.

cbeer:
I agree.. I've found a couple examples where maybe it'd make a difference:

```
EP 4.2:AR 2/DRAFT
EP 4.2:AR 2/2/DRAFT.
EP 4.2:AR 2/3/FINAL
EP 4.2:B 42/2/DRAFT/996
EP 4.2:B 44/2/998
```

but we'd probably break more things trying to figure out what is a word and what is just letters?

taylor-steve:
Hrm... Your example
```
EP 4.2:B 42/2/DRAFT/996
EP 4.2:B 44/2/998
```
is somewhat compelling to me. That could be more than an arm's length away... Maybe words are 4(?)+ characters in the suffix, with a precededing slash... That's kind of specific. I might need to sample more data and see if there are legit runs of letters that aren't words.

taylor-steve:
I guess the counter argument to that is "how are these actually being shelved"

cbeer:
... and how much trouble is it for the 6 call numbers where it'd make a difference?

... and are `NSIAD`, `JFMIP`, etc words or letters? or `DIR.`? or `SUM.`?

😬

taylor-steve:
I'm more than happy to exclude these edges cases for the moment and revisit as time/feedback justifies.

PR #1595, opened 2025-04-16
update public_note to look at holdings note for bound_withs

### LINKED ISSUE #1594: Treat bound-with child holdings notes as if they are public item notes
SW-4431

Per Jeanette:
> Upon migration bound-with children did not migrate items. Their Public item notes were migrated as Public Holdings notes, which do not display in SearchWorks.

dnoneill:
closes #1594

How this maps to SearchWorks

Screenshot 2025-04-16 at 6 18 56 PM

### DIFF
--- a/lib/folio_item.rb
+++ b/lib/folio_item.rb
@@ -66,7 +66,17 @@ class FolioItem
   end
 
   def public_note
-    @public_note ||= item&.dig('notes')&.map { |n| ".#{n['itemNoteTypeName']&.upcase}. #{n['note']}" }&.join("\n")&.presence
+    @public_note ||= (item_notes + (bound_with? ? holding_notes : []))
+
+    @public_note.map { |n| ".#{n[:type]&.upcase}. #{n[:note]}" }&.join("\n")&.presence
+  end
+
+  def item_notes
+    item&.dig('notes')&.map { |n| { type: n['itemNoteTypeName'], note: n['note'] } } || []
+  end
+
+  def holding_notes
+    holding&.dig('notes')&.map { |n| { type: n['holdingsNoteTypeName'], note: n['note'] } } || []
   end
 
   def call_number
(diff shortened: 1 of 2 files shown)


lib/folio_item.rb
```diff
@@ -66,7 +66,13 @@ def temporary_location_code
   end
 
   def public_note
-    @public_note ||= item&.dig('notes')&.map { |n| ".#{n['itemNoteTypeName']&.upcase}. #{n['note']}" }&.join("\n")&.presence
+    @public_note ||= get_notes(item&.dig('notes'), 'itemNoteTypeName')
+    @public_note += get_notes(holding&.dig('notes'), 'holdingsNoteTypeName') if bound_with?
+    @public_note&.join("\n")&.presence
```
cbeer, line 71: The `||=` doesn't seem quite right to me? won't that do funky stuff if we call `public_note` multiple times? I'd expect something more like this (but maybe there's a better refactoring):

```suggestion
    @public_note ||= begin
      notes = get_notes(item&.dig('notes'), 'itemNoteTypeName')
      notes += get_notes(holding&.dig('notes'), 'holdingsNoteTypeName') if bound_with?
      notes
    end  

    @public_note.join("\n").presence
```

---

I also wish I had a better suggestion for `get_notes` too; FOLIO's inconsistency makes me think too much. I don't know if extracting (say) `item_notes` and `holding_notes` would be cleaner (and separating the extraction from transforming it to a string)?

```
def public_note
  @public_note ||= (item_notes + (bound_with  ? holding_notes : []))

  @public_note.map { |n| ".#{n[:type]&.upcase}. #{n[:note]}" }&.join("\n")&.presence
end

def item_notes
  item&.dig('notes')&.map { |n| { type: n['itemNoteTypeName'], note: n['note'] } } || []
end

def holding_notes
  holding&.dig('notes')&.map { |n| { type: n['holdingsNoteTypeName'], note: n['note'] } } || []
end
```

🤷‍♂️

dnoneill: I think the second form makes more sense to me.

PR #1608, opened 2025-04-29
check indicator1=4 for 520 content advice vernacular fields

### LINKED ISSUE #1603: Only include unmatched vernacular for content advice if the ind1=4 (matching the regular field)
https://searchworks.stanford.edu/view/8567656 is displaying an unmatched vernacular 520 in both the contents/summary section and the content advice banner. It does not have a ind1=4, so it shouldn't be used in the content advice banner.

(SW-4435)

dnoneill:
closes #1603

### DIFF
--- a/lib/traject/common/marc_utils.rb
+++ b/lib/traject/common/marc_utils.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+CONTENT_ADVICE_LABEL = 'Content advice'
 module Traject
   module MarcUtils
     def extract_marc_and_prefer_non_alternate_scripts(spec, options = {})
@@ -346,7 +347,7 @@ module Traject
     end
 
     # @return [Array]
-    def get_unmatched_vernacular(marc, tag)
+    def get_unmatched_vernacular(marc, tag, label = '')
       return [] unless marc['880']
 
       fields = []
@@ -356,6 +357,9 @@ module Traject
         link = parse_linkage(field)
 
         next unless link[:number] == '00' && link[:tag] == tag
+
+        content_advice = (tag == '520' && field.indicator1 == '4')
+        next if content_advice != (label == CONTENT_ADVICE_LABEL)
 
         field.each do |sub|
           next if Constants::EXCLUDE_FIELDS.include?(sub.code)
--- a/spec/lib/traject/config/note_fields_spec.rb
+++ b/spec/lib/traject/config/note_fields_spec.rb
@@ -568,7 +568,7 @@ RSpec.describe 'Folio config' do
       expect(result.first[:fields].first[:field]).to include '520a', '520b'
     end
 
-    context 'with unmatched vernacular' do
+    context 'with unmatched vernacular for content advice' do
       let(:records) { [record] }
       let(:record) { nil }
       let(:result) { results.first }
@@ -578,7 +578,7 @@ RSpec.describe 'Folio config' do
         MARC::Record.new.tap do |r|
           r.append(
             MARC::DataField.new(
-              '880', '0', '0',
+              '880', indicator1, '0',
               MARC::Subfield.new('6', '520-00'),
               MARC::Subfield.new('a', '本书在综合考察绘画,文人,禅学这三者相互关联的基础上.')
             )
@@ -586,8 +586,23 @@ RSpec.describe 'Folio config' do
         end
       end
 
-      it 'maps the right fields' do
-        expect(result_field.first[:unmatched_vernacular]).to eq ['本书在综合考察绘画,文人,禅学这三者相互关联的基础上.']
+      context 'has 4 indicator' do
+        let(:indicator1) { '4' }
+
+        it 'maps the right fields' do
+          expect(result_field.length).to be 1
+          expect(result_field.first[:unmatched_vernacular]).to eq ['本书在综合考察绘画,文人,禅学这三者相互关联的基础上.']
+          expect(result_field.first[:label]).to eq 'Content advice'
+        end
+      end
+
+      context 'does not have a 4 indicator' do
+        let(:indicator1) { '0' }
+        it 'maps the right fields' do
+          expect(result_field.length).to be 1
+          expect(result_field.first[:unmatched_vernacular]).to eq ['本书在综合考察绘画,文人,禅学这三者相互关联的基础上.']
+          expect(result_field.last[:label]).to eq 'Summary'
+        end
       end
     end
 
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -102,7 +102,7 @@ Metrics/MethodLength:
 # Offense count: 3
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ModuleLength:
-  Max: 292
+  Max: 294
 
 # Offense count: 25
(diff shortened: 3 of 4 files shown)


lib/traject/common/marc_utils.rb
```diff
@@ -356,6 +356,7 @@ def get_unmatched_vernacular(marc, tag)
         link = parse_linkage(field)
 
         next unless link[:number] == '00' && link[:tag] == tag
+        next if tag == '520' && field.indicator1 != '4'
```
cbeer, line 359: The `summary_struct` has two cases:

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/folio_config.rb#L1541-L1544

The summary version should include unmatched vernacular that doesn't have the indicator1 = 4, and the content advice should include the unmatched vernacular  when indicator1=4. This breaks one of those, doesn't it?

cbeer: I'm not following.

Here's the example record today:

```
"summary_struct":[{"label":"Summary","fields":[],"unmatched_vernacular":["没什么不好理解的，雷神、鳄鱼、植物神的三者合一，这就是民间传说中的年。从神格上讲，年是雷神：从现实格上讲，年是鳄鱼；从功能格上讲，年是植物神。年就是这样既神秘又真实，成就了中国最隆重的传统节日--过年（春节）。这是一次在时间之路上对时间文化的探访。在中国时间文化的花园里，诸多时间性事物，特别是年、月、日这三朵金花，其不为人知的前世今生，其让人诧异的神性传说，其令人炫目的瑰丽愿景，都将一一呈现。-- From p. 4 of cover."]},{"label":"Content advice","fields":[],"unmatched_vernacular":["没什么不好理解的，雷神、鳄鱼、植物神的三者合一，这就是民间传说中的年。从神格上讲，年是雷神：从现实格上讲，年是鳄鱼；从功能格上讲，年是植物神。年就是这样既神秘又真实，成就了中国最隆重的传统节日--过年（春节）。这是一次在时间之路上对时间文化的探访。在中国时间文化的花园里，诸多时间性事物，特别是年、月、日这三朵金花，其不为人知的前世今生，其让人诧异的神性传说，其令人炫目的瑰丽愿景，都将一一呈现。-- From p. 4 of cover."]}],
```

The Content advice should not have an unmatched vernacular, because ind1 is not 4. It should look like:

```
"summary_struct":[{"label":"Summary","fields":[],"unmatched_vernacular":["没什么不好理解的，雷神、鳄鱼、植物神的三者合一，这就是民间传说中的年。从神格上讲，年是雷神：从现实格上讲，年是鳄鱼；从功能格上讲，年是植物神。年就是这样既神秘又真实，成就了中国最隆重的传统节日--过年（春节）。这是一次在时间之路上对时间文化的探访。在中国时间文化的花园里，诸多时间性事物，特别是年、月、日这三朵金花，其不为人知的前世今生，其让人诧异的神性传说，其令人炫目的瑰丽愿景，都将一一呈现。-- From p. 4 of cover."]}],
```

But if 880's ind1 was 4, it would be unmatched vernacular for content advice, e.g.:

```
"summary_struct":[{"label":"Content advice","fields":[],"unmatched_vernacular":["没什么不好理解的，雷神、鳄鱼、植物神的三者合一，这就是民间传说中的年。从神格上讲，年是雷神：从现实格上讲，年是鳄鱼；从功能格上讲，年是植物神。年就是这样既神秘又真实，成就了中国最隆重的传统节日--过年（春节）。这是一次在时间之路上对时间文化的探访。在中国时间文化的花园里，诸多时间性事物，特别是年、月、日这三朵金花，其不为人知的前世今生，其让人诧异的神性传说，其令人炫目的瑰丽愿景，都将一一呈现。-- From p. 4 of cover."]}],
```

I think this PR breaks one of those two, doesn't it?

dnoneill: Okay. I came up with a solution. I don't think it is particularly pretty but it works and shouldn't break anything

spec/lib/traject/config/note_fields_spec.rb
```diff
@@ -578,16 +578,31 @@
...
         end
       end
 
-      it 'maps the right fields' do
-        expect(result_field.first[:unmatched_vernacular]).to eq ['本书在综合考察绘画,文人,禅学这三者相互关联的基础上.']
+      context 'has 4 indicator' do
+        let(:indicator1) { '4' }
+
+        it 'maps the right fields' do
+          expect(result_field.length).to be 2
```
cbeer, line 593: I think there shouldn't be a Summary unmatched vernacular here?

```suggestion
          expect(result_field.length).to be 1
```

dnoneill: No it should be 2. When the indicator is 4 there should be a summary and content advice pair. When there isn't there should just be summary. Did I misunderstand the ticket again?

dnoneill: okay I think I understand what you are saying. You are saying 520 should only map to one field. Either Content Advice if the indicator is 4 and Summary if that isn't true?

cbeer: Exactly.

lib/traject/common/marc_utils.rb
```diff
@@ -357,6 +357,10 @@ def get_unmatched_vernacular(marc, tag)
 
         next unless link[:number] == '00' && link[:tag] == tag
 
+        content_advice = (tag == '520' && field.indicator1 == '4')
+        next if !content_advice && label == 'Content advice'
+        next if content_advice && label.include?('Summary')
```
taylor-steve, line 362: Are the labels the only thing we have to distinguish between these two situations? Not saying you're wrong here (my head is going in circles looking at all this), but I wonder if we can find a way where `get_unmatched_vernacular` doesn't depend on hardcoded label values. If this is the only way I wonder if we can at least push the burden of this knowledge out to the caller rather than embedding it here.

dnoneill: Not that I could think of. The function is called in https://github.com/sul-dlss/searchworks_traject_indexer/blob/9f1e6f5f5c199629a1a3cc53279134f846b43e6f/lib/traject/config/folio_config.rb#L1561 and https://github.com/sul-dlss/searchworks_traject_indexer/blob/9f1e6f5f5c199629a1a3cc53279134f846b43e6f/lib/traject/config/folio_config.rb#L1571. They do all the filtering so I am not quite sure how we would be able to distinguish which is a content advice and which is a summary. I guess we could create some contants but they would have to be shared across the folio_config and marc_utils. I think the instinct that these labels could change is good, I just don't know how to give a shared values across config and common

taylor-steve: I'm saying if this looks to be our only path forward maybe we consider pushing the label checking one level up so it's closer to where we define the label (I'm riffing this, untested):

```rb
def get_unmatched_vernacular(marc, tag, for_content_advice = false)
```

```rb
content_advice = (tag == '520' && field.indicator1 == '4')
next unless content_advice && for_content_advice
```

```rb
unmatched_vern = get_unmatched_vernacular(marc, tag, label == 'Content advice')
```

taylor-steve (approved):

PR #1609, opened 2025-04-29
fix 520 non text subfields index

dnoneill:
https://searchworks.stanford.edu/view/14289604 is indexing the c subfield. The a and b fields are the only content advice fields we are interested in https://www.loc.gov/marc/bibliographic/bd520.html. This stops the other subfields from being indexed.


cbeer:
gryphon-search is doing some analysis; per Kevin:

> there are 3000+ occurrences of 520 with 1st indicator '4'. ALL BUT 1 have the code "MiFhGG". No other codes exist. Depending on whether these 3000+ bibs will ever be replaced, it might make sense to just do a batch find/replace of that code with some appropriate statement like, "Provided by the Michigan Forestry Ham and Geese Group" (or just delete it completely if not important)
there are 5000+ occurrences of 520 with 2st indicator 0, 1, 2, 3, 8. Many of them give generic source statements like "provided by publisher" or "page 4 of cover"
that leaves 156K+ records that have 520 with 1st indicator is blank. Many/most are generic source statements.
there appears to only be 19 records in our catalog that have a 920$c. Almost all of these source notes seem to have been caused by an errant subfield delimiter in the MARC 920$a. Notice values like "olonial history", "ultural studies", "ommon knowledge", all missing the "c" on the first word. I think it would be safe to just fix up the MARC for these 19 bibs and ignore 920 completely.

saseestone:
In that gryphon-search analysis, I'm not seeing anything that would change our plans to only index/display the 520 indicator 4, subfields a/b. And stop indexing 520 indicator 4, subfield c. 

I'm also not seeing anything that would encourage us to change up how we handle other 520 indicators/subfields.  Does that seem right to you @cbeer ?

PR #1611, opened 2025-05-02
Remove building_location_facet_ssim

jcoyne:
This is not used in Searchworks


cbeer:
This is incredibly handy for finding records for testing + development. We shouldn't remove it.

PR #1612, opened 2025-05-02
Index SDR into the library_code_facet_ssim

jcoyne:
This allows us to use codes to display locations in SearchWorks

See https://github.com/sul-dlss/SearchWorks/pull/4858


cbeer:
#4858 references `library_code_facet_ssim`? Was that the field you meant to index?

jcoyne:
@cbeer yes. Thank you. Change made.

PR #1619, opened 2025-05-05
Use a predicate function for unmatched vernacular

taylor-steve:
@dnoneill what are your thoughts on something like this to avoid mixing presentation concerns with logic? Apologies if the draft PR is annoying, this was easier for me to riff with.

There's probably additional consolidation of logic for `summary` here, this is just the rough shape.


taylor-steve:
Yeah I'm thinking `accumulate_summary_struct_fields` and/or the nature of `matching_fields` needs a refactor and a cleaner solution will reveal itself

PR #1635, opened 2025-06-25
Add hierarchical field format_hsim to Folio indexing

### LINKED ISSUE #1632: Resource type facet indexing
In Searchworks 4.0, we're going to try to mush together resource types and formats into a single, hierarchical facet. The exact hierarchy may change, but the sooner we get the data structures available the better.

The SearchWorks implementation will end up looking something like:
Image

Here are the MARC mapping rules:
https://docs.google.com/document/d/1xR0cInOS2wlMzqGGY9ZLbxRb5PdUJ3b6jhnO6xr_upE/edit?tab=t.0

Presumably a similar list will (eventually) exist for MODS.

marlo-longley:
- Introduce `ResourceType` Traject macro module
- Define reusable matchers leader?, control_field_byte?, marc_subfield?
- Add literal_multiple method to Traject macro Extras

For reference see these files with the proposed changes and rules:
https://docs.google.com/document/d/1xR0cInOS2wlMzqGGY9ZLbxRb5PdUJ3b6jhnO6xr_upE/edit?tab=t.0
https://docs.google.com/spreadsheets/d/1NK5M1Ti8UG0XSfnRBWkgnF-seevfZWe3nEBowlV7R1g/edit?gid=0#gid=0

This PR does not cover every line, only the first few resource types. After the pattern is established and merged, I can continue down the list.
So this won't fully close #1632 but gets us partway there.

The macros take inspiration from `traject_plus` as well as the conditional logic included in Exhibits indexing.

### DIFF
--- a/lib/traject/config/folio_config.rb
+++ b/lib/traject/config/folio_config.rb
@@ -6,6 +6,7 @@ require 'csv'
 require 'i18n'
 require 'digest/md5'
 require 'active_support/core_ext/time'
+require_relative 'folio_format_config'
 
 I18n.available_locales = [:en]
 
@@ -13,6 +14,7 @@ extend Traject::Macros::Marc21
 extend Traject::Macros::Marc21Semantics
 extend Traject::SolrBetterJsonWriter::IndexerPatch
 extend Traject::MarcUtils
+extend FolioFormatConfig
 
 Utils.logger = logger
 
@@ -113,6 +115,9 @@ end
 def items(record, context)
   context.clipboard[:item] ||= record.index_items
 end
+
+# This handles the format_hsim fields in folio_format_config.rb
+add_folio_format_fields
 
 to_field 'id', extract_marc('001') do |_record, accumulator|
   accumulator.map! do |v|
--- a/lib/traject/config/folio_format_config.rb
+++ b/lib/traject/config/folio_format_config.rb
@@ -1,0 +1,139 @@
+# frozen_string_literal: true
+
+require_relative '../macros/folio_format'
+require_relative '../macros/extras'
+
+module FolioFormatConfig
+  include Traject::Macros::FolioFormat
+  include Traject::Macros::Extras
+
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
+  def add_folio_format_fields
+    to_field 'format_hsim',
+             condition(
+               leader?(byte: 6, values: %w[p t d f]),
+               literal('Archive/Manuscript')
+             )
+
+    to_field 'format_hsim',
+             all_conditions(
+               leader?(byte: 6, values: %w[a]),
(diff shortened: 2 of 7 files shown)


spec/lib/traject/config/resource_type_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+        context 'when 007[0] = g and 007[1] = s' do
+          let(:record) do
+            MARC::Record.new.tap do |r|
+              r.leader = '000000a00000000000000000'
+              r.append(MARC::ControlField.new('007', 'gs0000000000'))
+            end
+          end
+
+          it 'maps to Image|Slide' do
+            expect(result['resource_type_ssim']).to eq ['Image|Slide']
```
cbeer, line 260: For the hierarchy to work, we need both terms, e.g.:

```suggestion
            expect(result['resource_type_ssim']).to eq ['Image', 'Image|Slide']
```

lib/traject/config/folio_resource_type.rb
```diff
@@ -0,0 +1,140 @@
...
+
+require_relative '../macros/resource_type'
+
+module FolioResourceType
+  include Traject::Macros::ResourceType
+
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
+  def add_folio_resource_type_fields
+    to_field 'resource_type_ssim',
```
cbeer, line 11: This is going to drive a new facet field called "format", and it looks like our hierarchical call number field uses the suffix "hsim", so maybe something like 'format_hsim' is a better name? 

```suggestion
    to_field 'format_hsim',
```

lib/traject/config/folio_resource_type.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    to_field 'resource_type_ssim',
+             all_conditions(
+               leader?(byte: 6, values: %w[m]),
+               control_field_byte?('008', byte: 26, values: ['a']),
+               ->(_rec, acc, _ctx) { acc << 'Dataset' }
+             )
+
+    to_field 'resource_type_ssim' do |record, acc, _ctx|
+      logger.info(record.inspect)
```
cbeer, line 82:
```suggestion
```

lib/traject/macros/resource_type.rb
```diff
@@ -0,0 +1,88 @@
...
+            end
+          end
+        end
+      end
+
+      # Check a control field (like '008') at a specific byte for matching values
+      #
+      # Usage:
+      #   control_field_byte('008', byte: 26, values: ['a'])
+      def control_field_byte?(tag, byte:, values:)
```
cbeer, line 73: MARC 006, at least,  is a repeating control field (although I don't know how often that happens in practice?). Maybe we should iterate here, like you're doing for marc_subfield?

PR #1636, opened 2025-06-27
Add more fields to folio_format_config

marlo-longley:
With a few exceptions that will be clarified with librarians later in the week, I've added entries for rows 20-end in [the field spreadsheet](https://docs.google.com/spreadsheets/d/1NK5M1Ti8UG0XSfnRBWkgnF-seevfZWe3nEBowlV7R1g/edit?gid=0#gid=0).

### DIFF
--- a/lib/traject/config/folio_format_config.rb
+++ b/lib/traject/config/folio_format_config.rb
@@ -3,6 +3,7 @@
 require_relative '../macros/folio_format'
 require_relative '../macros/extras'
 
+# rubocop:disable Metrics/ModuleLength
 module FolioFormatConfig
   include Traject::Macros::FolioFormat
   include Traject::Macros::Extras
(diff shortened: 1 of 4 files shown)


lib/traject/config/folio_format_config.rb
```diff
@@ -109,31 +110,351 @@ def add_folio_format_fields
...
+             )
+
+    to_field 'format_hsim',
+             all_conditions(
+               control_field_byte?('007', byte: 0, value: 's'),
+               control_field_byte?('007', byte: 3, value: 'f'),
+               literal_multiple('Sound recording', 'Sound recording|CD')
+             )
+
+    piano_organ_roll_terms = ['piano roll', 'organ roll', 'audio roll']
```
cbeer, line 305: Is this checking if the 338a is exactly "piano roll", or contains "piano roll"?

PR #1647, opened 2025-07-11
Add object type indexing.

cbeer:


jcoyne (changes requested):
Should this have a test?

jcoyne (approved):

PR #1673, opened 2025-08-15
Refactor author `post_text` for a little more clarity.

cbeer:


spec/lib/traject/config/author_spec.rb
```diff
@@ -673,7 +673,7 @@
       expect(result).to include meeting: [{
         link: '111a none 111c 111d 111e none 111g 111i 111n 111q 111u',
         search: '111a none 111c 111d 111e none 111g 111i 111n 111q 111u',
-        post_text: '111e 111j'
```
cbeer, line 676: Confirmed with gryphon-search that `111$e` is something completely different from the `100$e` and `110$e` (which, we kinda knew in `#linked?`, I guess?).

taylor-steve (approved):

PR #1700, opened 2025-10-02
Drop unused field dor_read_rights_ssim

jcoyne:
The documentation said this could be used for finding fixtures of different types, but argo already serves this need. We also have an extensive list of fixtures documented here: https://github.com/sul-dlss/purl/blob/main/config/settings/development.yml\#L4-L47

This helps us decouple from XML metadata.


dnoneill:
I think https://github.com/sul-dlss/SearchWorks/blob/900944d6c422e7c72dc0049602aa4785d8c3c4e5/spec/fixtures/solr_documents/nj140cs3237.yml#L177-L178 probably should be removed first?

PR #1728, opened 2025-12-05
Remove PublicCocinaRecord seam

thatbudakguy:
This removes the `PublicCocinaRecord` class in favor of just accessing `CocinaDisplay::CocinaRecord` directly, since most of its methods are now just calling the corresponding one on `CocinaRecord`.

Adds a few shims to make this work; once #1726 and #1727 are done the shims can be removed since traject will call the appropriately named methods directly.

Also adds some Cocina versions of fixtures we only had MODS for, and stubs a few things in tests to get ready for everything pulling from Cocina.

I also switched the `#catkey` method to pull from Cocina instead of MODS; this resulted in different output in two tests but I think the new result is actually correct – the old version didn't have the `a` in the catkey, so this might actually result in more correct catkeys (which I know has occasionally been a problem).


spec/fixtures/files/df650pk4327.json
```diff
@@ -0,0 +1 @@
...liesTo":[]}],"metadataStandard":[],"identifier":[]},"purl":"https://purl.stanford.edu/df650pk4327"},"identification":{"catalogLinks":[],"sourceId":"hydrus:object-830"},"structural":{"contains":[{"type":"https://cocina.sul.stanford.edu/models/resources/document","externalIdentifier":"https://cocina.sul.stanford.edu/fileSet/df650pk4327-df650pk4327_1","label":"","version":2,"structural":{"contains":[{"type":"https://cocina.sul.stanford.edu/models/file","externalIdentifier":"https://cocina.sul.stanford.edu/file/df650pk4327-df650pk4327_1/2015ARCHIVING_IIIF.pdf","label":"","filename":"2015ARCHIVING_IIIF.pdf","size":1975323,"version":2,"hasMimeType":"application/pdf","sdrGeneratedText":false,"correctedForAccessibility":false,"hasMessageDigests":[{"type":"md5","digest":"23d199406499ab996fae3d4dcf08b21d"},{"type":"sha1","digest":"b27b109dfa6dc983b3bd2836d206292a1ef8d363"}],"access":{"view":"world","download":"world","controlledDigitalLending":false},"administrative":{"publish":true,"sdrPreserve":true,"shelve":true}}]}}],"hasMemberOrders":[],"isMemberOf":["druid:hn730ks3626"]},"created":"2025-07-07T13:17:40.000+00:00","modified":"2025-07-07T13:17:40.000+00:00","lock":"druid:df650pk4327=3=2"}
```
jcoyne: Why are we adding these fixtures in particular?  Can we add a manifest and describe their unique characteristics?

PR #1799, opened 2026-04-29
Don't index IIIF manifest URLs for GIS data

### LINKED ISSUE #1798: Don't index a IIIF manifest link for geo data content
This results in two problems:
- a largely useless manifest link is generated, since the item is not IIIF-viewable
- because the link is present, Geoblacklight (earthworks) will render a static map in the UI, which is at best duplication and at worst incorrect.

We should only index the manifest link for image/map/book items.

For example, qh585jp8667 (an index map) gets a IIIF manifest URL that points to an empty manifest, and [shows up in Earthworks](https://earthworks.stanford.edu/catalog/stanford-qh585jp8667) with a confusing static map that is unnecessary because we're already rendering it on top of a basemap:

Image

thatbudakguy:
Fixes #1798


jcoyne (commented):
I find that negative assertions are not very useful and add a lot of clutter.  Are they necessary?  Can we just add a unit test for iiif_manifest_url that says "geo" have none?

thatbudakguy:
Updated to do that.

jcoyne (approved):

PR #1805, opened 2026-05-12
Update logic for detecting georeferenced items

### LINKED ISSUE #1678: Improve logic behind georeferenced flag for Aardvark records
Right now this just looks for some defined words in the title: https://github.com/sul-dlss/searchworks_traject_indexer/blob/0ea81ddceb8134f958648987255563e810db54ad/lib/traject/config/geo_aardvark_config.rb#L178-L180

We can be a lot smarter.

An item is georeferenced if:
- it has "(Raster Image)" in the title (legacy geoTIFFs made from scanned maps)
- it is of type "map" or "image" and contains a .json file with the "georeference" role -> IIIF georeference annotations

It is not georeferenced if:
- it is of type "map" or "image" but does not contain a .json file with the "georeference" role

thatbudakguy:
This makes our georeference detection smarter, and apply to more
items, so that we can eventually add a georeference filter to the
Earthworks frontend if we want.

Fixes sul-dlss/earthworks#1678

### DIFF
--- a/lib/purl_record.rb
+++ b/lib/purl_record.rb
@@ -85,7 +85,7 @@ class PurlRecord
 
   delegate :collection?, :content_type, :files, :filesets, :cocina_doc, :world_access?,
            :modified_time, :created_time, :searchworks_url, :iiif_manifest_url,
-           :virtual_object?, :preferred_citation, :license, to: :public_cocina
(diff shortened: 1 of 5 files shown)


thatbudakguy:
Superseded by https://github.com/sul-dlss/earthworks/pull/1710.

PR #1829, opened 2026-08-04
Handle holdings records with multiple bound-with parents correctly

cbeer:
We've had a couple bug reports (SW-4620, SW-4686, SW-4687) recently about missing bound-with information, and I hope it all boils down to a mistake in how we're figuring out if an item is a bound-with parent. We've assumed a holdings record only has one bound-with (which is true for the bound-with child), but bound-with parents have a bound-with entry for each item.

I suspect there may be a better SQL query, but the internet seemed to think we'd have to add new, complex joins to get the right behavior... at least until we care to do more than flag the bound-with-ness of the principal, maybe this will do.


cbeer:
Alternatively, #1830 + #1831 are a PR stack for review.. if it works better this time.

PR #1830, opened 2026-08-04
Rename bound-with methods for a little more clarity

cbeer:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/lib/folio_item.rb
+++ b/lib/folio_item.rb
@@ -31,7 +31,7 @@ class FolioItem
   def initialize(item: nil, holding: nil, instance: nil,
                  course_reserves: [],
                  type: nil, status: nil,
-                 library: nil, record: nil, bound_with: false)
+                 library: nil, record: nil, bound_with_child: false)
     @item = item
     @holding = holding
     @instance = instance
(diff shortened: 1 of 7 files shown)


lib/folio_item.rb
```diff
@@ -165,13 +165,13 @@ def course_reserves_data
   end
 
   def bound_with
-    return unless bound_with?
+    return @bound_with if defined?(@bound_with)
 
-    holding&.dig('boundWith')
+    @bound_with ||= holding&.dig('boundWith') if @bound_with_child
```
taylor-steve, line 170: Maybe?
```suggestion
    return unless @bound_with_child

    @bound_with ||= holding&.dig('boundWith')
```

taylor-steve (approved):

lib/folio_item.rb
cbeer: I'm not sure. The previous implementation of `bound_with?` checked what we sent in, so I kept it here.

There's some screwy things with how things are modeled (the bound-with principal's holdings record also has a `boundWith` entry pointing at the principal)

I'm not sure it's worth too much analysis, because I clean up some of the weirdness later in the PR stack.

PR #1835, opened 2026-08-06
Display volume-only items without indexing them as call numbers

taylor-steve:
For items like https://searchworks.stanford.edu/view/L222 that currently display `(no call number)`, we noticed Requests uses the volume information which seems like nicer behavior for the user.

Before:
Screenshot 2026-08-06 at 10 08 32 AM

After:
Screenshot 2026-08-06 at 10 08 16 AM

This requires a fair number of spec guards to protect ourselves. I think it's okay-ish, but wouldn't be upset if folks thought we needed to explore refactors first (e.g., split out some concept of `indexable?`, `browseable?`, `displayable?` rather than leaning on ignored).

### DIFF
--- a/lib/folio_item.rb
+++ b/lib/folio_item.rb
@@ -198,14 +198,16 @@ class FolioItem
                          @holding&.dig('callNumber') ||
                          bound_with&.dig('holding', 'callNumber')
 
-    return CallNumber.new('', '') unless base_call_number.present?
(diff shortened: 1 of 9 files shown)


jcoyne (approved):

jcoyne:
@taylor-steve needs a rebase.

PR #1837, opened 2026-08-06
Extract bound-with query to get it out of holding

cbeer:
We're currently  getting duplicate holdings records when there are multiple bound-with principals associated with a holdings record.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/lib/folio/holding.rb
+++ b/lib/folio/holding.rb
@@ -2,21 +2,26 @@
 
 module Folio
   class Holding
-    attr_reader :data, :pieces, :items
+    attr_reader :data, :pieces, :items, :bound_with_principal
 
     delegate :[], :fetch, :dig, :as_json, :to_json, to: :data
 
-    def self.from_dynamic(holding, items: [], pieces: [])
+    def self.from_dynamic(holding, items: [], pieces: [], bound_with_principals: [], bound_with_parts: [])
       this_pieces = pieces.select { |p| p['holdingId'] == holding['id'] }
       this_items = items.select { |i| i['holdingsRecordId'] == holding['id'] }
+      bound_with_link = bound_with_parts.find { |p| p['holdingsRecordId'] == holding['id'] }
+      bound_with_principal = bound_with_principals.find { |p| p.dig('item', 'id') == bound_with_link['itemId'] } if bound_with_link
 
-      new(holding, pieces: this_pieces, items: this_items)
+      new(holding, pieces: this_pieces, items: this_items, bound_with_principal: bound_with_principal)
     end
 
-    def initialize(data, pieces: [], items: [])
+    def initialize(data, pieces: [], items: [], bound_with_principal: nil)
       @data = data
+      # keep the bound-with data in the original data structure (for now) to avoid breaking existings tests + code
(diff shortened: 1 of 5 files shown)


lib/folio/holding.rb
```diff
@@ -2,21 +2,25 @@
...
-    attr_reader :data, :pieces, :items
+    attr_reader :data, :pieces, :items, :bound_with_principal
 
     delegate :[], :fetch, :dig, :as_json, :to_json, to: :data
 
-    def self.from_dynamic(holding, items: [], pieces: [])
+    def self.from_dynamic(holding, items: [], pieces: [], bound_with_principals: [])
       this_pieces = pieces.select { |p| p['holdingId'] == holding['id'] }
       this_items = items.select { |i| i['holdingsRecordId'] == holding['id'] }
+      bound_with_principal = bound_with_principals.find { |p| p.dig('holding', 'id') == holding['id'] }
```
taylor-steve: I'm looking at a failure to index https://searchworks.stanford.edu/view/4244926.

I'm having zero fun trying to trace this 😁 I might be missing something, bound withs are hard...

I think drilling into `bound_with_principals[].holding.id` comes from `parentHolding.id` via the SQL query currently?

`holding['id']` here is the child holding being wrapped? So with the exception of the the self-bound (SW-4330) case, they will never match, so the match returns nil, `@data['boundWith']` is never merged, and `Holding#bound_with?` returns false for the child. 

So when I'm indexing, `bound_with_holdings_as_stub_items` produces nothing for the child records.

I haven't gotten quite far enough yet, but do we need another clause in the `bound_with_principals` aggregation to pull in the child hr.id?

cbeer: hm. I'm not seeing an issue and end up with:

```
{"id":"d2b0230e-5975-570a-b467-83fa63e2838a","barcode":"36105008325016","library":"SAL3","type":"book","note":null,"instance_id":"5e5ca7bf-0346-5e11-af6d-f15ce102aaa5","instance_hrid":"a4244926","effective_permanent_location_code":"SAL3-STACKS","temporary_location_code":null,"permanent_location_code":"SAL3-STACKS","status":"Available","effective_location_id":"1146c4fa-5798-40e1-9b8e-92ee4c9f2ee2","material_type_id":"1a54b431-2e4f-452d-9cae-9cee66c9a892","loan_type_id":"2b94c631-fca9-4892-a730-03ee529ffe27","bound_with":null,"is_bound_with_principal":true,"lopped_callnumber":"595.706 .C15","callnumber":"595.706 .C15 V.11 1956-1957","full_shelfkey":"dewey 595.70600000 c15 v.000011 001956-001957"}
```

taylor-steve: Hrm, what am I doing wrong?

main:
Screenshot 2026-08-07 at 8 50 59 AM

this branch:
Screenshot 2026-08-07 at 8 52 13 AM

taylor-steve (approved):

